PR #26, opened 2018-01-22
MQE-661

KevinBKozan:
Break out common test actions from di.xml

### Description
All test actions that are common between before/after/test/action-group are now an entity declared at the top of di.xml. If actions are added to this list, they now can only be added to one list instead of six.

### Fixed Issues
1. MQE-661: Break out common test actions from di.xml

### Contribution checklist
 - [x] All automated tests passed successfully (all builds on Travis CI are green)


KevinBKozan:
I will squash and merge to get rid of all of these commits.

PR #76, opened 2018-04-02
Added the ability to use entity properties in metadata

rocketweb:
### Description

Added the ability to use entity properties inside metadata entities for when you make API requests to create something that has to contain an array of properties from other entities, e.g. the request for creating a new tax rule has to contain an array of the ids of the tax rates that will be associated to it.

Now you can have a meta file that represents a tax rate, which would look something like this:

````
<?xml version="1.0" encoding="UTF-8"?>
<operations xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:noNamespaceSchemaLocation="../../../../../../vendor/magento/magento2-functional-testing-framework/src/Magento/FunctionalTestingFramework/DataGenerator/etc/dataOperation.xsd">
    <operation name="CreateTaxRate" dataType="tax_rate" type="create" auth="adminOauth" url="/V1/taxRates" method="POST">
        <contentType>application/json</contentType>
        <object dataType="tax_rate" key="taxRate">
            <field key="code">string</field>
            <field key="tax_country_id">string</field>
            <field key="tax_region_id">integer</field>
            <field key="tax_postcode">string</field>
            <field key="rate">integer</field>
        </object>
    </operation>
</operations>
````
And a meta file that represents a tax rule, which would look something like this:

````
<?xml version="1.0" encoding="UTF-8"?>
<operations xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:noNamespaceSchemaLocation="../../../../../../vendor/magento/magento2-functional-testing-framework/src/Magento/FunctionalTestingFramework/DataGenerator/etc/dataOperation.xsd">
    <operation name="CreateTaxRule" dataType="tax_rule" type="create" auth="adminOauth" url="/V1/taxRules" method="POST">
        <contentType>application/json</contentType>
        <object dataType="tax_rule" key="rule">
            <field key="code">string</field>
            <field key="priority">integer</field>
            <field key="position">integer</field>
            <array key="customer_tax_class_ids">
                <value>integer</value>
            </array>
            <array key="product_tax_class_ids">
                <value>integer</value>
            </array>
            <array key="tax_rate_ids">
                <value>tax_rate.id</value>
            </array>
        </object>
    </operation>
</operations>
````

Notice the `<value>tax_rate.id</value>` inside the `tax_rule` entity. Now a tax rule can be created successfully.

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/DataGenerator/Persist/OperationDataArrayResolver.php
+++ b/src/Magento/FunctionalTestingFramework/DataGenerator/Persist/OperationDataArrayResolver.php
@@ -134,6 +134,14 @@ class OperationDataArrayResolver
                     ));
                 }
             } else {
+
+                $operationElementProperty = null;
+                if(strpos($operationElementType, '.') !== false){
+                    $operationElementComponents = explode('.',$operationElementType);
+                    $operationElementType = $operationElementComponents[0];
+                    $operationElementProperty = $operationElementComponents[1];
+                }
+
                 $entityNamesOfType = $entityObject->getLinkedEntitiesOfType($operationElementType);
 
                 // If an element is required by metadata, but was not provided in the entity, throw an exception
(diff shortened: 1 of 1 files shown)


KevinBKozan:
@rocketweb,
Following up with you about the license/CLA. If you could please sign the license/cla we can process this Pull Request.

PR #89, opened 2018-04-06
Add ability to use array entities as arguments.

nathanjosiah:
Added support for passing array entities as arguments.

### Description

As a Test Engineer I want to pass data entity arrays as arguments so that I can not have to use hard-coded flat arrays.

For example:

**The Entity**

```xml
<entities>
    <entity name="DefaultCustomerGroup" type="customerGroup">
        <array key="group_names">
            <item>Default</item>
        </array>
    </entity>
</entities>
```

**The ActionGroup**

```xml
<actionGroups>
    <actionGroup name="searchAndMultiSelectActionGroup">
        <arguments>
            <argument name="options" />
        </arguments>
        <selectMultipleOptions filterSelector="foo" optionSelector="bar" stepKey="selectSpecifiedOptions">
            <array>{{options}}</array>
        </selectMultipleOptions>
    </actionGroup>
</actionGroups>
```

**The Test**

```xml
<tests>
    <test name="AdminCreateCustomerTest">
        <actionGroup ref="searchAndMultiSelectActionGroup" stepKey="searchAndSelectGroup">
            <argument name="options" value="DefaultCustomerGroup.group_names"/>
        </actionGroup>
    </test>
</tests>
```

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
+++ b/src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
@@ -495,6 +495,10 @@ class ActionObject
                 $this->setTimeout($obj->getElement($objField)->getTimeout());
             } elseif (get_class($obj) == EntityDataObject::class) {
                 $replacement = $this->resolveEntityDataObjectReference($obj, $match);
+
+                if (is_array($replacement)) {
+                    $replacement = '["' . implode('","', array_map('addSlashes', $replacement)) . '"]';
+                }
             }
 
             if ($replacement === null) {
@@ -503,9 +507,6 @@ class ActionObject
                 } else {
                     throw new TestReferenceException("Could not resolve entity reference " . $inputString);
                 }
-            }
-            elseif (is_array($replacement)) {
-                $replacement = '["' . implode('","', array_map('addSlashes',$replacement)) . '"]';
             }
 
             $replacement = $this->resolveParameterization($parameterized, $replacement, $match, $obj);
--- a/src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
+++ b/src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
@@ -496,6 +496,10 @@ class ActionObject
 
             $replacement = $this->resolveParameterization($parameterized, $replacement, $match, $obj);
 
+            if (is_array($replacement)) {
+                $replacement = '["' . implode('","', $replacement) . '"]';
+            }
+
             $outputString = str_replace($match, $replacement, $outputString);
         }
         return $outputString;
--- a/src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
+++ b/src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
@@ -493,12 +493,11 @@ class ActionObject
                     throw new TestReferenceException("Could not resolve entity reference " . $inputString);
                 }
             }
+            elseif (is_array($replacement)) {
+                $replacement = '["' . implode('","', array_map('addSlashes',$replacement)) . '"]';
+            }
 
             $replacement = $this->resolveParameterization($parameterized, $replacement, $match, $obj);
-
-            if (is_array($replacement)) {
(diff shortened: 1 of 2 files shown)


src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
```diff
@@ -496,6 +496,10 @@ private function findAndReplaceReferences($objectHandler, $inputString)
 
             $replacement = $this->resolveParameterization($parameterized, $replacement, $match, $obj);
 
+            if (is_array($replacement)) {
+                $replacement = '["' . implode('","', $replacement) . '"]';
```
nathanjosiah: I'm aware this would be insufficient escapement but I wasn't sure how far this needed to go. I figured if nothing else this PR serves as a proof of concept and opens the conversation to this feature.

KevinBKozan: Hey @nathanjosiah, what do you mean that it's `insufficient escapement`? Are you talking about how the escapement assumes the array's items are `string`s?

KevinBKozan: @nathanjosiah, by `insufficient escapement` do you mean how it assumes that the array members are `string`s?

nathanjosiah: @KevinBKozan Unless multidimensional arrays have a use case then I'm not thinking as much about that. I was thinking more about if the items contain quotes or `$`. The resulting Codception unit test would potentially be malformed or inadvertently contain PHP variable interpolation.

KevinBKozan: Gotcha. Let me leave some review comments on some small stuff then!

KevinBKozan: We should array_walk/array_map with `addSlashes` to break quotes at the very least. Inadvertend variable interpolation in MFTF is not yet handled, so this is not introducing anything new.

src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
```diff
@@ -496,6 +496,10 @@ private function findAndReplaceReferences($objectHandler, $inputString)
 
             $replacement = $this->resolveParameterization($parameterized, $replacement, $match, $obj);
 
+            if (is_array($replacement)) {
```
KevinBKozan: This check should either live in the elseif block in 497, or in the `resolveEntityDataObjectReference()` function. This conditional only ever applies to EntityDataObjects, so checking it every time is unecessary.

nathanjosiah:
@KevinBKozan I updated my branch with the requested changed and merged mainline and ran tests to verify current compatibility.

src/Magento/FunctionalTestingFramework/Test/Objects/ActionObject.php
```diff
@@ -504,6 +504,9 @@ private function findAndReplaceReferences($objectHandler, $inputString)
                     throw new TestReferenceException("Could not resolve entity reference " . $inputString);
                 }
             }
+            elseif (is_array($replacement)) {
+                $replacement = '["' . implode('","', array_map('addSlashes',$replacement)) . '"]';
```
KevinBKozan: @nathanjosiah my bad, I don't think I was very clear about this portion. I meant that your addition should be entirely inside the `elseif` on line 496; that `elseif` is the portion that deals with the object passed in being an `EntityDataObject`. Since the only time `$replacement` could only ever be an `array` is if it's an array inside an `EntityDataObject`, we should only every have to check it inside that conditional.

Does that make sense?

nathanjosiah: Ah, yes I see what you mean. I have pushed an updated.

nathanjosiah:
@KevinBKozan I have updated the code and the Travis build is now passing.

KevinBKozan:
@nathanjosiah Awesome. Code additions look good and all the builds are passing, thank you very much for the contribution!

PR #218, opened 2018-08-30
MQE-1236

KevinBKozan:
### Description
- Changed AnnotationExtractor to automatically prepend `testCaseId` to the `title` annotation, and added a check for uniqueness of `testCaseId` to `title` annotation.
- All of this is to prevent Allure from overwriting (calling duplicate titles `retries`) 

2. ...

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Test/Util/AnnotationExtractor.php
+++ b/src/Magento/FunctionalTestingFramework/Test/Util/AnnotationExtractor.php
@@ -22,6 +22,11 @@ class AnnotationExtractor extends BaseObjectExtractor
      */
     private $storyToTitleMappings = [];
 
+    /**
+     * @var array
+     */
+    private $testCaseToTitleMappings = [];
+
     const ANNOTATION_VALUE = 'value';
     const MAGENTO_TO_ALLURE_SEVERITY_MAP = [
         "BLOCKER" => "BLOCKER",
@@ -90,8 +95,8 @@ class AnnotationExtractor extends BaseObjectExtractor
             $annotationObjects[$annotationKey] = $annotationValues;
         }
 
+        $this->addTestCaseIdToTitle($annotationObjects, $filename);
         $this->validateMissingAnnotations($annotationObjects, $filename);
-
         $this->addStoryTitleToMap($annotationObjects, $filename);
 
         return $annotationObjects;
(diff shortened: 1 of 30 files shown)


src/Magento/FunctionalTestingFramework/Test/Util/AnnotationExtractor.php
```diff
@@ -162,6 +191,32 @@ public function validateStoryTitleUniqueness()
...
+                $dupes[$newTitle] = "'" . implode("', '", $files) . "'";
+            }
+        }
+        if (!empty($dupes)) {
+            $message = "TestCaseId and Title pairs must be unique:\n\n";
+            foreach ($dupes as $newTitle => $tests) {
+                $testCaseTitleArray = explode(" - ", $newTitle);
+                $testCaseId = $testCaseTitleArray[0];
+                $title = $testCaseTitleArray[1];
+                $message .= "TestCaseId: '{$testCaseId}' Title: '{$title}' in Tests {$tests}\n\n";
```
jilu1: Maybe you want to tweak the error message a bit when $testCaseId is "[No TestCaseID]"? This is when there is a dupe where both don't have test case id and share the same title.

src/Magento/FunctionalTestingFramework/Test/Util/AnnotationExtractor.php
```diff
@@ -112,6 +117,30 @@ public function addStoryTitleToMap($annotations, $filename)
...
+            return;
+        }
+
+        $testCaseId = "[NO TESTCASEID]";
+
+        if (isset($annotations['testCaseId'])) {
+            $testCaseId = $annotations['testCaseId'][0];
+        }
+
+        $newTitle = "{$testCaseId} - " . $annotations['title'][0];
```
jilu1: This is total optional. would it be better to use ":" instead of "-" to concatenate test case ID and title, because there is a "-" inside test case ID and our commit message is formatted like "MAGETWO-123456: ".

KevinBKozan: Good point, I will do this.

src/Magento/FunctionalTestingFramework/Test/Util/AnnotationExtractor.php
KevinBKozan: I actually kind of like the exception message:
`TestCaseId: '[NO TESTCASEID]' Title: 'Magento Functional Testing Framework Documentation is available.' in Tests 'DevDocsTest', 'DevDocsTest2'`
The `[NO TESTCASEID]` is that gets appended to the title (meaning it propagates all the way to the allure report). I kind of like that consistency, but let me know if you feel strongly about it.

jilu1: let's fine.

PR #243, opened 2018-10-03
MQE-1279: MFTF will not output any test results if instance respondes…

jilu1:
… with 503

- fix error handling to simplify stack trace

2. ...

### Contribution checklist


src/Magento/FunctionalTestingFramework/Module/MagentoRestDriver.php
```diff
@@ -96,25 +97,31 @@ class MagentoRestDriver extends REST
      */
     public function _beforeSuite($settings = [])
     {
-        parent::_beforeSuite($settings);
-        if (empty($this->config['url']) || empty($this->config['username']) || empty($this->config['password'])) {
-            return;
-        }
+        try {
```
KevinBKozan: Changes don't really help the root issue. In the build, this will still cause a silent failure, as this is run before the entire suite (or test), meaning no failure artifacts are recorded at all.

I tried removing this file from `functional.suite.yml` and found that it solves the issue entirely; the magento instance is unavailable but the test starts properly and fails naturally due to the instance being in maintenance mode.
Can we remove the reference to `MagentoRestDriver` from `functional.suite.yml` entirely? I think this is old unused code.

jilu1: I disabled `MagentoRestDriver`, run functional test builds and it looks fine. 

Although it does not make sense for silent failure in Jenkins as an exception is thrown either from Codeception or MFTF framework. Still Jenkins pipeline issue?

KevinBKozan: Jenkins funs `codecept run` statements with `|| true`, meaning it never uses the codeception exit code (or the `MagentoRestDriver` exception in this case) to determine failure, it only uses existance of test artifacts.

PR #251, opened 2018-10-11
MQE-1260: Create RERUN_COUNT field in Jenkins MFTF Parameters section same as MTF parameters section

aljcalandra:
Adding a run:failed command to run the tests that failed on the last build

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
+++ b/src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
@@ -1,0 +1,145 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types = 1);
+
+namespace Magento\FunctionalTestingFramework\Console;
+
+use Magento\FunctionalTestingFramework\Suite\Handlers\SuiteObjectHandler;
+use Magento\FunctionalTestingFramework\Config\MftfApplicationConfig;
+use Magento\FunctionalTestingFramework\Test\Handlers\TestObjectHandler;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Process\Process;
+use Magento\FunctionalTestingFramework\Exceptions\TestFrameworkException;
+
+class RunTestFailedCommand extends BaseGenerateCommand
+{
+    const DEFAULT_TEST_GROUP = 'default';
+
+    /**
+     * Configures the current command.
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        $this->setName('run:failed')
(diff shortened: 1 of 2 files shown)


src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
```diff
@@ -0,0 +1,190 @@
...
+            );
+        }
+    }
+
+    /**
+     * Returns a json string of tests that failed on the last run
+     *
+     * @return string
+     */
+    private function getFailedTestList()
```
KevinBKozan: Rerunning `failed` tests twice does not meet expected behaviour. This is what happens:
* Run each test/suite via `codecept run` and keep track of `failed` tests (like the build will)
* Run `run:failed`, all of the failed tests are reran
* Run `run:failed` again, only the last failed `codecept run` statement is re-ran (since the `failed` file has been overwritten in the middle of the `mftf` command). I think you need to keep track of the contents of that `failed` file in between these `codecept run` statements, which is what I think the `rerun_tests` file is for.

aljcalandra: Alrighty, I'll work on this.
The rerun failed file won't work though since we are tracking the tests that are rerun, not the rerun tests that failed.

src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
```diff
@@ -0,0 +1,192 @@
...
+            $process->setIdleTimeout(600);
+            $process->setTimeout(0);
+            $process->run(
+                function ($type, $buffer) use ($output) {
+                    $output->write($buffer);
+                }
+            );
+            $this->failedList = array_merge(
+                $this->failedList,
+                $this->readFailedTestFile(self::TESTS_FAILED_FILE)
```
KevinBKozan: Found while testing, if your last `codecept run` statement contains no failures, it actually deletes the `failed` file, so if a flaky test passes on the second try, `$this->readFailedTestfile` will actually fail to open the stream to the file (which was deleted).

src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
```diff
@@ -0,0 +1,194 @@
...
+            $process = new Process($codeceptionCommand);
+            $process->setWorkingDirectory(TESTS_BP);
+            $process->setIdleTimeout(600);
+            $process->setTimeout(0);
+            $process->run(
+                function ($type, $buffer) use ($output) {
+                    $output->write($buffer);
+                }
+            );
+            if (realpath(self::TESTS_FAILED_FILE)) {
```
KevinBKozan: For some reason, this `realpath` still allows the conditional to pass if the fiel has been deleted.

Locally, I've found that adding `file_get_contents()` to the `readFailedTestFile` function works to prevent the nulling out of `$this->failedTests`:
```private function readFailedTestFile($filePath)
{
    if (file_get_contents($filePath) !== false) {
        return file($filePath, FILE_IGNORE_NEW_LINES);
    }
    return [];
}```

KevinBKozan:
@aljcalandra Rerun looks good ot go from a local perspective. If you end up needing to make more changes to this due to jenkins implementation, I will happily re-review.

PR #252, opened 2018-10-16
MQE-1152: Remove redundant after execution for codeception 2.4.4 and update composer.json

aljcalandra:
Added check for codeception version to prevent running after statement twice in later versions of codeception

### Contribution checklist


src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
```diff
@@ -115,13 +116,15 @@ private function runAfterBlock($e, $cest)
...
             $actorClass = $e->getTest()->getMetadata()->getCurrent('actor');
             $I = new $actorClass($cest->getScenario());
-            call_user_func(\Closure::bind(
-                function () use ($cest, $I) {
-                    $cest->executeHook($I, 'after');
-                },
-                null,
-                $cest
-            ));
+            if (version_compare(Codecept::VERSION, TestContextExtension::CODECEPT_AFTER_VERSION, "<=")) {
```
KevinBKozan: You need an explicit `use Codeception\Codecept;` for PHP7.0 compatibility.

pmclain:
Are there plans to merge this? The current codeception version restraint locks Magento on PHPUnit 6.

KevinBKozan:
@pmclain Yes this PR is going to be merged. Only issue with it is some decoupling work required in the Magento2 codebase that is essentially an artifact of when MFTF was embedded in `dev/tests/acceptance`.
The `bin/mftf` commands in Magento2CE call the root codeception (which in this case will likely be `2.4.0`) but sets CWD to `dev/tests/acceptance`. Codeception does this thing that it redirects to use a version installed under `<cwd>/vendor/bin/codeception` if it finds it, which in this case will be `2.3.6`. This causes the process to crash due to some weird compatibility problems from switching the version on the fly, hence the need for decoupling work.

My gut tells me we will only be able to merge this after the next Magento patch release, as that's the only way to include the decoupling work.

KevinBKozan:
@pmclain So we can merge this work, but I don't think it will solve your issues with phpunit versioning. We cannot version bump to Codeception 2.4.x until we remove Codeception from `dev/tests/acceptance` in Magento; since that is a Magento codebase change, the codeception version bump will need to wait for the next Magento release (which I assume would be `2.3.1` patch).

Let me know if the above makes sense; it's a sticky circular dependency issue that can be solved, just not until the change can make it onto the Magento codebase.

pmclain:
@KevinBKozan Thanks, that's perfect. I was hoping to update the phpunit version in upstream magento, but saw the version was actually restricted by codeception. I'll keep an eye on this PR and look out for an updated MFTF version in 2.3-develop.

PR #258, opened 2018-10-26
Removed unused variables in FunctionCommentSniff.php

gy741:
### Description

Hello,

I have remove unused variable.

Because it is not used.

What do you think of it?

Thanks.

N/A

### Contribution checklist


KevinBKozan:
@gy741 Thank you for your contribution! We will update this PR with progress information as we process it.

dev/tests/static/Magento/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -460,7 +460,7 @@ protected function processParams(File $phpcsFile, $stackPtr, $commentStart)
                     $phpcsFile->fixer->replaceToken(($param['tag'] + 2), $content);
 
                     // Fix up the indent of additional comment lines.
-                    foreach ($param['commentLines'] as $lineNum => $line) {
+                    foreach ($param['commentLines'] as $lineNum) {
```
okolesnyk: I believe you should not do this change.
Because if we have  `foreach ($array as $key => $value)`  in this case $lineNum is a $key, but if we use like you suggested `foreach ($array as $value)`  $lineNum is going to be a $value. which is incorrect.

PR #269, opened 2018-11-05
Spaceship operator

lbajsarowicz:
### Description
I've changed the ArrayType interpreter for MFTF. By the way created draft of PHPUnit Tests, as well as added some argument type hints and return type hints

### Contribution checklist

### DIFF
--- a/dev/tests/phpunit.xml
+++ b/dev/tests/phpunit.xml
@@ -21,6 +21,7 @@
     </testsuites>
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="false">
+            <directory suffix=".php">../../src/Magento/FunctionalTestingFramework/Data</directory>
             <directory suffix=".php">../../src/Magento/FunctionalTestingFramework/DataGenerator</directory>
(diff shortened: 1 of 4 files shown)


aljcalandra:
@lbajsarowicz This technically changes functionality since in the case of them being equal we would return 1, while your change would return 0 in that case.

I'll review to see if this affects code outside of this area. If you already know or are aware just let us know.

lbajsarowicz:
@aljcalandra just briefly: I can't imagine well composed code which has the same values with the same key to be compared, and that's the case you are asking about.

lbajsarowicz:
@okolesnyk  - It's not being used for returning `-1` or `1` for specific reason, but for comparing items as a callback for `uksort()`

https://github.com/M2Coach/magento2-functional-testing-framework/blob/6d3085640930acbf8e9177df99ef3a5fdfe7d46d/src/Magento/FunctionalTestingFramework/Data/Argument/Interpreter/ArrayType.php#L64-L69

PR #272, opened 2018-11-12
MQE-1350: Add support for symlinked vendor/magento modules

KevinBKozan:
### Description
- Added clause for checking current symlinked module
- This can only be reproduced by setting `repositories` in the composer file when installing Magento extensions via composer. 
Example flow is
- git clone extension repo
- add local path to repository ```{ "type": "path", "url":
        "/<myExtension>/app/code/Magento/<Module>" }```
- composer require <myExtension>
Above leads to a symlinked module under `magento/vendor/magento/module-<module>/`, which causes the moduleResolver to not be able to resolve the symlink.

### Contribution checklist


KevinBKozan:
@jilu1 I've simplified the fix such that we are resolving the path we check for
``` <path> / * / <pattern>```

PR #314, opened 2019-02-27
MQE-810: Create a static test to validate references between modules

KevinBKozan:
### Description
* Implemented `static-checks`
* Checks the following types of files:
   * Test.xml
   * Data.xml
   * ActionGroup.xml
* Performs checks on following reference types:
   * xml entity references like`{{data.xml}}` and `{{data.xml(also.parameters)}}`
   * `extends=""`
   * actionGroup `ref=""`
* Validation is based finding common depdency in file's module and **any of the filenames from MFTF entity**
   * Above means if `dataEntity` has base or merge in `catalog` and `wishlist`, reference will be valid if file has dendepncy on either `catalog` or `wishlist`.

### Contribution checklist


src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -0,0 +1,385 @@
...
+                    }
+                }
+                if (!$valid) {
+                    $violatingReferences[$entityName] = $files;
+                }
+            }
+
+            if (!empty($violatingReferences)) {
+                // Build error output
+                $errorOutput = "\nFile \"{$filePath->getRealPath()}\"\ncontains references to following modules:\n\t\t";
```
jilu1: Should we point out the violation directly? e.g. using words like "SomeTest should not contain references to SomeModule" would be better?

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -0,0 +1,385 @@
...
+        $outputPath = getcwd() . DIRECTORY_SEPARATOR . "mftf-dependency-checks.txt";
+        $fileResource = fopen($outputPath, 'w');
+        $header = "MFTF File Dependency Check:\n";
+        fwrite($fileResource, $header);
+        foreach ($errors as $test => $error) {
+            fwrite($fileResource, $error[0] . PHP_EOL);
+        }
+        fclose($fileResource);
+        $errorCount = count($errors);
+        $output = "Dependency errors found across {$errorCount} file(s). Error details output to {$outputPath}";
```
jilu1: Is it number total tests scanned or number of total errors found? Just look at the error message in console confused me until I see the source code.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
KevinBKozan: I can change the error output to be more precise, yes.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
KevinBKozan: Errors unique to file, the `$error` array is key => value of `fileName => [$errors]` so the count just counts unique file names.

KevinBKozan:
@jilu1 I'm not sure how merge file contents play in here, files are being checked on a per-file basis, not per entity (concept of merging is removed).
I reran my tests and am not finding that the above files are shown as errors. Did you pull down the right branch? It's `MQE-810-B`, due to the extensive refactor.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -0,0 +1,385 @@
...
+        $registrar = new \Magento\Framework\Component\ComponentRegistrar();
+        $this->moduleNameToPath = $registrar->getPaths(\Magento\Framework\Component\ComponentRegistrar::MODULE);
+        $this->moduleNameToComposerName = $this->buildModuleNameToComposerName($this->moduleNameToPath);
+        $this->flattenedDependencies = $this->buildComposerDependencyList();
+
+        $allModules = ModuleResolver::getInstance()->getModulesPath();
+        $filePaths = [
+            DIRECTORY_SEPARATOR . 'Test' . DIRECTORY_SEPARATOR,
+            DIRECTORY_SEPARATOR . 'ActionGroup' . DIRECTORY_SEPARATOR,
+            DIRECTORY_SEPARATOR . 'Data' . DIRECTORY_SEPARATOR,
```
jilu1: Why not include Page and Section here?

jilu1:
> @jilu1 I'm not sure how merge file contents play in here, files are being checked on a per-file basis, not per entity (concept of merging is removed).
> I reran my tests and am not finding that the above files are shown as errors. Did you pull down the right branch? It's `MQE-810-B`, due to the extensive refactor.

You are right. I was using the wrong branch. Please ignore the previous comment

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
KevinBKozan: Didn't think it was relevant, mostly. Section can't reference anything other than static selector strings, and while Pages can reference Sections, that linkage is not really used anywhere in MFTF.
If we do build an actual use for Page->Section linkage I think we should add that link check, but otherwise I don't see it as value adding as far as the check goes

jilu1: I see

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -0,0 +1,385 @@
...
+    private function findErrorsInFileSet($files)
+    {
+        $testErrors = [];
+        foreach ($files as $filePath) {
+            $modulePath = dirname(dirname(dirname(dirname($filePath))));
+            $moduleFullName = array_search($modulePath, $this->moduleNameToPath) ?? null;
+            // Not a module, is either dev/tests/acceptance or loose folder with test materials
+            if ($moduleFullName == null) {
+                continue;
+            }
```
jilu1: I am running standalone mftf, and I have a test module under mftf's dev/tests/functional/tests/MFTF/CatalogTest which has been scanned but it should not.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -0,0 +1,385 @@
...
+        }
+        return $flattenedDependencies;
+    }
+
+    /**
+     * Recursive function to fetch dependencies of given dependency, and its child dependencies
+     * @param string $subDependencyName
+     * @return array
+     */
+    private function extractSubDependencies($subDependencyName)
```
jilu1: This recursive function is fine by itself. But any reason that you don't build the recursion from root level composer file. Will the current way cause duplicate dependencies in the root level?

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -0,0 +1,385 @@
...
+                foreach ($splitArguments as $argument) {
+                    // Do nothing for 'string' or $persisted.data$
+                    if (preg_match(ActionObject::STRING_PARAMETER_REGEX, $argument)) {
+                        continue;
+                    } elseif (preg_match(TestGenerator::PERSISTED_OBJECT_NOTATION_REGEX, $argument)) {
+                        continue;
+                    }
+                    // trim `data.field` to `data`
+                    preg_match('/([^.]+)/', $argument, $entityName);
+                    // Double check that {{data.field}} isn't an argument for an ActionGroup
```
jilu1: Why skipping? What if it's a direct entity reference in an argument for an ActionGroup? e.g. {{_defatultCategory.name}}

KevinBKozan: Skipping done on line 175 refers to the following case:
```
<argument name="product" defaultValue="_defaultProduct"/>
...
<comment userInput="{{product.name}} />
<comment userInput="{{_defaultProduct.name}} />
```

The entity `product` is just the argument, and if we have a match of `name` we shouldn't try to resolve it, as the entity doesn't exist.

So in this case the first line would be skipped, but line two would not be.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
KevinBKozan: I **think** the reason is because it would flatten all the dependencies to one huge array, and we want an array with several `module => [dependencies]`.
We iterate through the big list of dependencies on ln 283 so that we can scope the above flattening.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
KevinBKozan: Will check/resolve this.

src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
jilu1: ok. This is right. I did not look the regex and just thought the argument you tried to skip is in the "value" part.

PR #320, opened 2019-03-21
MQE-1224: Remove fillSecretField from Test Schema

tomreece:
### Description
* Removed fillSecretField from test schema

### Contribution checklist


tomreece:
I believe this change is backwards compatible and does not need to wait for a major version.

The fillSecretField action could not have be used without throwing an error message (see https://github.com/magento/magento2-functional-testing-framework/blob/develop/src/Magento/FunctionalTestingFramework/Test/Util/ActionMergeUtil.php#L112)

KevinBKozan:
@tomreece You are correct for the most part.
So since per-file schema validation isn't run by default, you are correct in that it explicitly `does not break generation`. You **can** use `<fillSecretField>` just like you can use `<whateveractionhere>`, the generator does not whitelist actions to an explicit list, the schema does that validation.

There isn't a good reason for anyone to be using `fillSecretField` due to what you pointed out, and evern if someone had created a test using it, normal generation isn't broken.

PR #324, opened 2019-03-29
MQE-1376: [SPIKE] Investigate Self-Documentation for MFTF

aljcalandra:
Adding a command `generate:docs` to allow for local generation of documentation for action groups

### Description
Adding unit tests, documentation and a command to MFTF to enable self-documentation.

### Contribution checklist


src/Magento/FunctionalTestingFramework/Util/DocGenerator.php
```diff
@@ -0,0 +1,193 @@
...
+     * @var ActionGroupObjectHandler
+     */
+    private static $instance;
+
+    /**
+     * Array of action groups
+     *
+     * @var ActionGroupObjectHandler
+     */
+    private $actionGroups;
```
KevinBKozan: Unused local variable, you can remove this.

src/Magento/FunctionalTestingFramework/Util/DocGenerator.php
```diff
@@ -0,0 +1,193 @@
...
+    const ANNOTATION_DESCRIPTION = "description";
+    const FILENAMES = "filenames";
+    const ARGUMENTS = "arguments";
+
+    /**
+     * Single instance of class var
+     *
+     * @var ActionGroupObjectHandler
+     */
+    private static $instance;
```
KevinBKozan: Does this need to be a singleton as opposed to a static class that you could call `DocGenerator::generateDocs()`? The only public function you have in this class is passed in all it needs to generate the docs, and as far as I can tell this class does not store any data for long term use (which imo defeats the purpose of the singleton).

src/Magento/FunctionalTestingFramework/Util/DocGenerator.php
```diff
@@ -0,0 +1,193 @@
...
+    const DEFAULT_OUTPUT_DIR =
+        PROJECT_ROOT .
+        DIRECTORY_SEPARATOR .
+        "dev" .
+        DIRECTORY_SEPARATOR .
+        "tests" .
+        DIRECTORY_SEPARATOR .
+        "docs";
+    const DOC_NAME = "documentation.md";
+    const ANNOTATION_MODULE = "";
```
KevinBKozan: Leftover const I believe

src/Magento/FunctionalTestingFramework/Util/DocGenerator.php
```diff
@@ -0,0 +1,193 @@
...
+    const ANNOTATION_PAGE = "page";
+    const ANNOTATION_DESCRIPTION = "description";
```
KevinBKozan: Does it make sense to keep these `const` in `ActionGroupObject` instead? I'm just thinking of maintenance cost if we change schema or add more annotations, where does it make sense to keep these `const`s that correspond to schema changes.

dev/tests/_bootstrap.php
```diff
@@ -60,6 +60,28 @@
 defined('TESTS_BP') || define('TESTS_BP', __DIR__);
 defined('TESTS_MODULE_PATH') || define('TESTS_MODULE_PATH', TESTS_BP . $RELATIVE_TESTS_MODULE_PATH);
 defined('MAGENTO_BP') || define('MAGENTO_BP', __DIR__);
+define('OUTPUT_DIR',
```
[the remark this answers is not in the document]
aljcalandra: 👍

src/Magento/FunctionalTestingFramework/Console/GenerateDocsCommand.php
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Configures the current command.
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        $this->setName('generate:docs')
+            ->setDescription('This command generates documentation for created MFTF files.')
+            ->addOption(
```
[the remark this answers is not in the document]
aljcalandra: I thought about it, I guess I was considering this as a separate ticket, but it would be simple to do here.

src/Magento/FunctionalTestingFramework/Test/etc/actionGroupSchema.xsd
```diff
@@ -30,6 +30,14 @@
                     </xs:sequence>
                 </xs:complexType>
             </xs:element>
+            <xs:element name="annotations">
```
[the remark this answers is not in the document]
aljcalandra: This would be backwards incompatible right?

src/Magento/FunctionalTestingFramework/Util/DocGenerator.php
aljcalandra: I'm not sure on this one. this seems to be different per area of the framework.

src/Magento/FunctionalTestingFramework/Test/etc/actionGroupSchema.xsd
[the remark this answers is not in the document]
KevinBKozan: @aljcalandra Yes, mostly. We don't run schema validation by default on builds/etc, but if we changed that this would be breaking.
I'll create a ticket to make these required in the backlog for later implementation.

src/Magento/FunctionalTestingFramework/Util/DocGenerator.php
KevinBKozan: Well, mentally I think `actionGroup schema` => `actionGroup class` if I needed to take a look at any constants. Similarly in `actionObject` we have `const ACTION_ATTRIBUTE_TIMEZONE = 'timezone'` for the attribute that lives in the schema.

I don't feel all that strongly on this one though, if you disagree then feel free to ignore this.

src/Magento/FunctionalTestingFramework/Console/GenerateDocsCommand.php
```diff
@@ -0,0 +1,79 @@
...
+            MftfApplicationConfig::GENERATION_PHASE,
+            false,
+            false
+        );
+
+        $allActionGroups = ActionGroupObjectHandler::getInstance()->getAllObjects();
+        $docGenerator = new DocGenerator();
+        $docGenerator->createDocumentation($allActionGroups, $config, $clean);
+
+        $output->writeln("Generate Docs Command Run");
```
KevinBKozan: I would print the output directory here like `Generated to xxx`, so it's clear where exactly it was generated to.

PR #354, opened 2019-05-23
MQE-659: [ALLURE] Include the test "stepKey" in the MFTF Allure Report

jilu1:

2. ...

### Contribution checklist


KevinBKozan:
@jilu1 I'm not sure that the StepKeyReader class is 100% necessary; in the `stepBefore` hook, you can get access to the step's line number via `$stepEvent->getStep()`. That way we could do something really simple:
* Keep read file in memory (in testBefore or something)
* In `stepBefore` hook, read line from file in memory and strip out the `//stepKey: <key>`
* Append to step name there

Since we could strip out the line itself we don't have to worry about parsing the file at all, thus eliminating the need for the reader class. Let me know if the suggestion makes sense, I feel pretty strongly in keeping code additions as trim as possible!

PR #386, opened 2019-07-18
MQE-1606: Mftf static check script should return error exit code when check fails

tomreece:
### Description
The static-checks command now returns 1 if any errors are found and 0 if there are no errors.

https://jira.corp.magento.com/browse/MQE-1606

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
+++ b/src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
@@ -6,20 +6,20 @@
 
 namespace Magento\FunctionalTestingFramework\StaticCheck;
 
-use Magento\FunctionalTestingFramework\Config\Data;
 use Magento\FunctionalTestingFramework\Config\MftfApplicationConfig;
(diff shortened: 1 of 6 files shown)


src/Magento/FunctionalTestingFramework/StaticCheck/TestDependencyCheck.php
```diff
@@ -62,6 +62,12 @@ class TestDependencyCheck implements StaticCheckInterface
      */
     private $alreadyExtractedDependencies;
 
+    /**
+     * Array containing all errors found after running the execute() function.
+     * @var array
+     */
+    private $errors;
```
KevinBKozan: This new variable reveals that the original deisgn of the `TestDependencyCheck` should have been a little different.

* `execute()` should not return anything
* Class should have another new variable `private $output` so it can be accessible like `$staticCheck->getOutput()`

This is more future static check friendly, as it opens up the window for more nuanced error tracking/etc.

Keep in mind, you'll need to also edit the `StaticCheckInterface` but it's a simple edit.

tomreece:
@KevinBKozan -- I acted on your feedback and refactored a few minor things along the way. Let me know if I made things better or worse.

PR #396, opened 2019-07-26
MQE-1672: AllureAdapter does not output BROKEN test steps

KevinBKozan:
### Description
- Fixed broken steps not showing up at all
- Fixed ActionGroup nesting issues when the test abruptly ends

### Contribution checklist


src/Magento/FunctionalTestingFramework/Allure/Adapter/MagentoAllureAdapter.php
```diff
@@ -187,10 +189,37 @@ public function testIncomplete(FailEvent $failEvent)
...
+    {
+        $event = new TestCaseBrokenEvent();
+        $e = $failEvent->getFail();
+        $message = $e->getMessage();
+
+        // Create new step with an error for Allure
+        $failStep = new Step();
+        $failStep->setName("ERROR");
+        $failStep->setTitle($message);
+        $failStep->setStatus(Status::BROKEN);
```
jilu1: Is the BROKEN status accurate?

KevinBKozan: I think so, yes. Class `Status` only has mappings for `FAILED BROKEN PASSED CANCELED PENDING`, so I think it makes sense.

Also, refer to `AllureCodeception.php` ln 339, you'll see two functions `testError()` and `testFail()`. 
*`testError()` => hard error in test, like a fatal PHP error
* `testFail()` => exception thrown in the test cause a true failure (not an error)

Based on that delinitation and mapping to Codeception's existing hooks I think it makes sense for this to be `Status::BROKEN`

jilu1: That makes sense.

PR #429, opened 2019-08-23
MQE-1582: Enable Testers To Run Skipped Tests

KevinBKozan:
## Description
- added `--allow-skipped` flag

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Config/MftfApplicationConfig.php
+++ b/src/Magento/FunctionalTestingFramework/Config/MftfApplicationConfig.php
@@ -9,6 +9,9 @@ use Magento\FunctionalTestingFramework\Exceptions\TestFrameworkException;
 
 class MftfApplicationConfig
 {
+    /**
+     * MFTF Execution Phases
+     */
     const GENERATION_PHASE = "generation";
     const EXECUTION_PHASE = "execution";
     const UNIT_TEST_PHASE = "testing";
@@ -51,6 +54,12 @@ class MftfApplicationConfig
     private $debugLevel;
 
     /**
+     * Boolean which allows MFTF to fully generate skipped tests
+     * @var bool
+     */
+    private $allowSkipped;
+
+    /**
      * MftfApplicationConfig Singelton Instance
      *
      * @var MftfApplicationConfig
@@ -64,13 +73,15 @@ class MftfApplicationConfig
      * @param string  $phase
      * @param boolean $verboseEnabled
      * @param string  $debugLevel
+     * @param boolean $allowSkipped
      * @throws TestFrameworkException
      */
     private function __construct(
         $forceGenerate = false,
         $phase = self::EXECUTION_PHASE,
         $verboseEnabled = null,
-        $debugLevel = self::LEVEL_NONE
+        $debugLevel = self::LEVEL_NONE,
+        $allowSkipped = false
     ) {
         $this->forceGenerate = $forceGenerate;
 
@@ -89,6 +100,7 @@ class MftfApplicationConfig
             default:
                 $this->debugLevel = self::LEVEL_DEVELOPER;
         }
+        $this->allowSkipped = $allowSkipped;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/Magento/FunctionalTestingFramework/Config/MftfApplicationConfig.php
```diff
@@ -157,6 +172,16 @@ public function getDebugLevel()
...
     }
 
+    /**
+     * Returns a boolean indicating whether mftf is generating skipped tests.
+     *
+     * @return boolean
+     */
+    public function allowSkipped()
+    {
+        return $this->allowSkipped ?? getenv('ALLOW_SKIPPED');
```
jilu1: Does mftf or codeception support ALLOW_SKIPPED?

src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
```diff
@@ -28,6 +29,23 @@ protected function configure()
             'r',
             InputOption::VALUE_NONE,
             'remove previous generated suites and tests'
+        )->addOption(
+            "force",
+            'f',
+            InputOption::VALUE_NONE,
+            'force generation of tests regardless of Magento Instance Configuration'
```
jilu1: How about reword it a bit to include "run test", something similar like below for "allowSkip"?

src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
```diff
@@ -28,6 +29,23 @@ protected function configure()
...
+            'a',
+            InputOption::VALUE_NONE,
+            'Allows MFTF to generate and run skipped tests.'
+        )->addOption(
+            'debug',
+            'd',
+            InputOption::VALUE_OPTIONAL,
+            'Run extra validation when running tests. Use option \'none\' to turn off debugging -- 
+             added for backward compatibility, will be removed in the next MAJOR release',
+            MftfApplicationConfig::LEVEL_DEFAULT
```
jilu1: Similar like previous comment. mention both "generate" and "run".

src/Magento/FunctionalTestingFramework/Console/GenerateTestsCommand.php
```diff
@@ -128,18 +117,26 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-    private function createTestConfiguration($json, array $tests, bool $force, $debug, bool $verbose)
-    {
+    private function createTestConfiguration(
+        $json,
+        array $tests,
+        bool $force,
+        $debug,
+        bool $verbose,
+        bool $allowSkipped
+    ) {
```
jilu1: Can we either use "type param" or "param"? e.g. $json and $debug don't have type but the other params use type.

src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
```diff
@@ -76,13 +76,17 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        $force = $input->getOption('force');
```
jilu1: Line 54 ~ 61 are already defined in base class, and should be removed.

src/Magento/FunctionalTestingFramework/Config/MftfApplicationConfig.php
KevinBKozan: MFTF, codeception does not have a notion of `skipped` tests at all.

src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
KevinBKozan: Good point, will do 👍

src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
KevinBKozan: Will do 👍

src/Magento/FunctionalTestingFramework/Console/GenerateTestsCommand.php
KevinBKozan: I was pattern matching to use `type param`, I can change the ones that don't have it to do the same.

I think this is a static check that we don't have enabled in our framework, @okolesnyk prefers type hinting at the signature so I'm going down that route.

src/Magento/FunctionalTestingFramework/Console/RunTestFailedCommand.php
KevinBKozan: Good catch, will fix.

src/Magento/FunctionalTestingFramework/Config/MftfApplicationConfig.php
jilu1: Then we introduce this ALLOW_SKIPPED? in .env?

KevinBKozan: Good point, I was pattern matching other configs. I'll add this to the default + docs

PR #434, opened 2019-08-28
MQE-1703: Implicit Suite Generation for Tests

soumyau:

2. ...

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
+++ b/src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
@@ -70,7 +70,7 @@ class BaseGenerateCommand extends Command
     }
 
     /**
-     * Returns a 2D array of tests with their suites references that can be encoded into a json test configuration
+     * Returns an array of test configuration to be used as an argument for generation of tests
      * @param array $tests
      * @return false|string
      * @throws \Magento\FunctionalTestingFramework\Exceptions\XmlException
--- a/src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
+++ b/src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
@@ -81,24 +81,23 @@ class BaseGenerateCommand extends Command
         $testConfiguration['tests'] = null;
         $testConfiguration['suites'] = null;
         $testsReferencedInSuites = SuiteObjectHandler::getInstance()->getAllTestReferences();
-        $resolvedTests = [];
+        $suiteToTestPair = [];
 
         foreach($tests as $test) {
             if (array_key_exists($test, $testsReferencedInSuites)) {
                 $suites = $testsReferencedInSuites[$test];
-                $resolvedTests = array_merge(
-                    $resolvedTests,
-                    array_map(function ($value) use ($test) {
-                        return  $value . ':' . $test;
-                    }, $suites)
-                );
+                foreach ($suites as $suite) {
+                    $suiteToTestPair[] = "$suite:$test";
+                }
             }
             // configuration for tests
-            else $testConfiguration['tests'][] = $test;
+            else {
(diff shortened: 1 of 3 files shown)


src/Magento/FunctionalTestingFramework/Console/GenerateTestsCommand.php
```diff
@@ -65,14 +65,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $json = $input->getOption('tests'); // for backward compatibility
         $force = $input->getOption('force');
         $time = $input->getOption('time') * 60 * 1000; // convert from minutes to milliseconds
         $debug = $input->getOption('debug') ?? MftfApplicationConfig::LEVEL_DEVELOPER; // for backward compatibility
         $remove = $input->getOption('remove');
         $verbose = $output->isVerbose();
         $allowSkipped = $input->getOption('allowSkipped');
 
+        if (!empty($tests)) {
+            $json = $this->getTestAndSuiteConfiguration($tests);
```
KevinBKozan: There's quite a bit of weird little implications of making this change (I don't disagree with the change though).

Due to the way TestGenerator uses a local `$this-tests` variable for some reason, and due to the fact that we are now *always* passing in something to that variable, I would suggest the following:

* Pass in the *entire* testsConfiguration in `ln 96`
* Run through TestGenerator's uses of `$this->tests` and refactor use so that it understands the new config being passed in
* Unit test the following usecases
** Test0 (no suite) Test1 (suite1)
** generate:tests (everything is generated)
** generate:tests Test0 (only Test0 is generated)
** generate:tests Test1 (only Test1 is generated)

I started off seeing incorrect behavior with testcase 3 which led me down a rabbit hole of playing whackamole with changes, and I think you will just need to deep dive into TestGenerator and do refactoring of how it uses the `$this->tests` variable.

soumyau: Put in a fix by initializing the json configuration for `tests` and `suites` to null instead of [] which generates/runs all tests/suites in `getTestAndSuiteConfiguration(array $tests)`

src/Magento/FunctionalTestingFramework/Console/RunTestCommand.php
```diff
@@ -104,4 +111,27 @@ function ($type, $buffer) use ($output) {
...
+     */
+    private function resolveSuiteReferences($testConfigurationJson)
+    {
+        $testConfiguration = json_decode($testConfigurationJson, true);
+        $testsArray = $testConfiguration['tests'] ?? [];
+        $suitesArray = $testConfiguration['suites'] ?? [];
+        $testArrayBuilder = [];
+
+        foreach ($suitesArray as $suite => $tests) {
+            $testArrayBuilder = array_merge(
```
KevinBKozan: See commend in BaseGenerateCommand, this can probably be simplified into one `foreach`

src/Magento/FunctionalTestingFramework/Console/BaseGenerateCommand.php
```diff
@@ -67,4 +68,40 @@ protected function removeGeneratedDirectory(OutputInterface $output, bool $verbo
...
+    {
+        $testConfiguration['tests'] = null;
+        $testConfiguration['suites'] = null;
+        $testsReferencedInSuites = SuiteObjectHandler::getInstance()->getAllTestReferences();
+        $resolvedTests = [];
+
+        foreach($tests as $test) {
+            if (array_key_exists($test, $testsReferencedInSuites)) {
+                $suites = $testsReferencedInSuites[$test];
+                $resolvedTests = array_merge(
```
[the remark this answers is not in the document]
soumyau: was trying to avoid nested loops as I didn't find many implementations of it in code, thought it was a style thing. Changed it.

PR #436, opened 2019-08-29
MQE-1633 + MQE-1514

KevinBKozan:
## Description
- Added AllureHelper class + unit tests
- Added calls to above class in MagentoWebDriver `see/dontsee` url actions + `CurlHandler` for persistence actions

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Allure/AllureHelper.php
+++ b/src/Magento/FunctionalTestingFramework/Allure/AllureHelper.php
@@ -23,7 +23,8 @@ class AllureHelper
     }
 
     /**
-     * Adds Attachment to the last executed step. Required due to Allure root-step behavior
+     * Adds Attachment to the last executed step.
+     * Use this when adding attachments outside of an $I->doSomething() step/context.
      * @param mixed  $data
      * @param string $caption
      * @return void
--- a/src/Magento/FunctionalTestingFramework/Allure/AllureHelper.php
+++ b/src/Magento/FunctionalTestingFramework/Allure/AllureHelper.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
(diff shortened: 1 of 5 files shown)


src/Magento/FunctionalTestingFramework/Allure/AllureHelper.php
```diff
@@ -0,0 +1,39 @@
...
+        Allure::lifecycle()->fire(new AddAttachmentEvent($data, $caption));
+    }
+
+    /**
+     * Adds Attachment to the last executed step. Required due to Allure root-step behavior
+     * @param mixed  $data
+     * @param string $caption
+     * @return void
+     */
+    public static function addAttachmentToLastStep($data, $caption)
```
tomreece: Just curious about in general when to use this over the current step? I see you use this one in the curl handler, but I cant figure out why. Maybe add a bit to the phpdoc like:

`This function should be used when ___blahblahblah___`

tomreece:
I do not see the request/response data in the build: http://10.234.204.237:8080/job/Functional-Tests-CE/2021/allure/#suites/8187718e2fd144987673ef34ee62b41b/136f33b4de6d9efe/

Am I blind?

KevinBKozan:
@tomreece I forgot to commit a file for unit tests, that will fix the errors.

And let me retrigger/reset up branches for the build, I may have messed something up there.

src/Magento/FunctionalTestingFramework/Allure/AllureHelper.php
KevinBKozan: I'll add a comment to clear it up, the gist of it is that during a step (aka a function called in the test via `$I->doathing()`) you use current. If the framework is doign its own logic outisde a function like that, you can add artifact to the last step (which is usually a comment step)

tomreece:
Looks good to me.

PR #437, opened 2019-08-29
MQE-1626: bin/mftf run:manifest

tomreece:
### Description
I created a new run:manifest command which is able to run testManifest.txt files.
testManifest.txt files are created when tests are generated or can be crafted by hand to meet the user's needs.

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Console/RunManifestCommand.php
+++ b/src/Magento/FunctionalTestingFramework/Console/RunManifestCommand.php
@@ -7,6 +7,7 @@ declare(strict_types = 1);
 
 namespace Magento\FunctionalTestingFramework\Console;
 
+use Magento\FunctionalTestingFramework\Exceptions\TestFrameworkException;
 use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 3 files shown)


src/Magento/FunctionalTestingFramework/Console/RunManifestCommand.php
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Executes the run:manifest command.
+     *
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     * @return integer
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $manifestFile = file($input->getArgument("path"), FILE_IGNORE_NEW_LINES);
```
soumyau: if path/to/your/testManifest.txt is invalid it throws a failed to open steam exception, how about adding an explicit error checking,
`if (!realpath($input->getArgument("path")))`

tomreece: I added an explicit check for the input file.

PR #445, opened 2019-09-05
Mqe 1650 Update MFTF configuration to read Test entities from new location

jilu1:

2. ...

### Contribution checklist


src/Magento/FunctionalTestingFramework/Page/Config/Dom.php
```diff
@@ -87,13 +87,13 @@ public function initDom($xml, $filename = null)
...
-            '_' .
-            $this->modulePathExtractor->getExtensionPath($filename);
+            $this->modulePathExtractor->getExtensionPath($filename)
+            . '_'
+            . $this->modulePathExtractor->extractModuleName($filename);
         foreach ($pageNodes as $pageNode) {
             $pageModule = $pageNode->getAttribute("module");
             $pageName = $pageNode->getAttribute("name");
-            if ($pageModule !== $currentModule) {
+            if ($pageModule != $currentModule) {
```
okolesnyk: Please try to keep strict comparison.

src/Magento/FunctionalTestingFramework/Test/Util/TestObjectExtractor.php
```diff
@@ -11,6 +11,7 @@
 use Magento\FunctionalTestingFramework\Test\Objects\ActionObject;
 use Magento\FunctionalTestingFramework\Test\Objects\TestObject;
 use Magento\FunctionalTestingFramework\Util\ModulePathExtractor;
+use Magento\FunctionalTestingFramework\Util\ModuleResolver;
```
okolesnyk: Is it necessary to add this `use` here?

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
```diff
@@ -0,0 +1,130 @@
...
+            throw new TestFrameworkException("Invalid directory: {$directory}");
+        }
+
+        // Find all composer json files under directory
+        $modules = [];
+        $jsonFinder = new ComposerJsonFinder();
+        $fileList = $jsonFinder->finComposerJsonFilesAtDepth($normalizedDir, 2);
+        foreach ($fileList as $file) {
+            // Parse composer json for test module name and path information
+            $composerInfo = new ComposerPackager(
```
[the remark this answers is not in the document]
KevinBKozan: Additionally if you are able to remove `ComposerPackager`, I'd get rid of the `AbstractComposerHandler` class as well. As far as I can tell it'd be a collection of getters and a single util function (that can be defined elsewhere).

etc/config/.env.example
```diff
@@ -43,7 +43,7 @@ BROWSER=chrome
 #DEFAULT_TIMEZONE=America/Los_Angeles
 
 #*** These properties impact the modules loaded into MFTF, you can point to your own full path, or a custom set of modules located with the core set
-MODULE_WHITELIST=Magento_Framework,Magento_ConfigurableProductWishlist,Magento_ConfigurableProductCatalogSearch
+MODULE_WHITELIST=Magento_Framework,ConfigurableProductWishlist,ConfigurableProductCatalogSearch
```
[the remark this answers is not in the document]
jilu1: This are plain white list module paths. They should not be assumed to be magento module to match based on current logic

src/Magento/FunctionalTestingFramework/Page/Config/Dom.php
jilu1: Done

src/Magento/FunctionalTestingFramework/Test/Util/TestObjectExtractor.php
jilu1: Code has changed and it's not needed any more. Removed.

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -179,50 +247,84 @@ public function getEnabledModules()
...
      * Return the modules path based on which modules are enabled in the target Magento instance.
      *
+     * @param boolean $verbose
      * @return array
      */
-    public function getModulesPath()
+    public function getModulesPath($verbose = false)
     {
-        if (isset($this->enabledModulePaths)) {
+        if (isset($this->enabledModulePaths) && !$verbose) {
```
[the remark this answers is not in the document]
jilu1: The $verbose variable here does not mean framework verbosity. It's defined just in getModulesPath(). Let me know if you want me to change to a different name.

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -179,50 +247,84 @@ public function getEnabledModules()
...
             return $this->enabledModulePaths;
         }
 
+        if (isset($this->enabledModuleNameAndPaths) && $verbose) {
```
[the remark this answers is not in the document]
jilu1: $verbose is a local variable.

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -314,46 +429,193 @@ private static function globRelevantWrapper($testPath, $pattern)
...
-     * @param array $modulePaths
      * @return array
      */
-    private function flattenAllModulePaths($modulePaths)
+    private function aggregateTestModulePathsFromComposerJson()
     {
-        $it = new \RecursiveIteratorIterator(new \RecursiveArrayIterator($modulePaths));
-        $resultArray = [];
+        // Define the module paths
+        $magentoBaseCodePath = MAGENTO_BP;
```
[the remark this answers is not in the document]
jilu1: This is a copy/paste from existing code in line 334. It makes the code more meaningful. Let me know if you still suggest to change in all places?

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -314,46 +429,193 @@ private static function globRelevantWrapper($testPath, $pattern)
...
-            } elseif (isset($this->knownDirectories[$magentoModuleName])
-                && !isset($allModulePaths[$magentoModuleName])) {
-                LoggingUtil::getInstance()->getLogger(ModuleResolver::class)->warn(
-                    "Known directory could not match to an existing path.",
-                    ['knownDirectory' => $magentoModuleName]
-                );
+        if (null !== $this->composerJsonModulePaths) {
+            return $this->composerJsonModulePaths;
+        }
+        try {
```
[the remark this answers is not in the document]
jilu1: getTestModules() could throw TestFrameworkException when directory passed in is invalid. It will be in logs but here we ignore this exception and continue execution.

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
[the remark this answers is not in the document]
jilu1: The ComposerPackager class is an object to represent composer.json file. We will work with it a lot in the future, like module dependencies, packaging etc. We might need to add new functions to it. It makes sense to me to be a class itself.

Another consideration is ModuleResolver class is already 880+ lines of code. We actually should consider refactor it to make it more object oriented and shorter but not adding more utility functions into it.

After review the ComposerPackager and AbstractComposerHandler, I have moved parseSuggestsForMagentoModuleNames() in to ComposerPackager where it more belongs.

AbstractComposerHandler is to hold common composer factory property for both classes.

Thoughts?

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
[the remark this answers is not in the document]
KevinBKozan: If you think it helps code verbosity, then no worries 👍

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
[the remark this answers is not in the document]
KevinBKozan: Gotcha.

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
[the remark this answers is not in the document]
KevinBKozan: Gotcha, please rename it then. I thought that this meant `--verbose` at first.

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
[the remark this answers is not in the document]
KevinBKozan: Hm. I see the use in having a `ComposerPackage` (without `r` at end) but I hesitate to add classes until we have a clearly defined need for it, especially since the PR already introduces so many new classes.

I agree with keeping it out of ModuleResolver though, I was thinking you could add the util function to `ComposerModuleResolver` which is still at ~120 lines (if we do end up removing the ComposerPackager class).

I would ask you to just try the following and see the result:
* Remove `ComposerPackager` and `AbstractComposerHandler` and port any required logic to `ComposerModuleResolver` + `ComposerInstaller`. 
* Rename `ComposerInstaller` to `ComposerHandler`

My main goal in any of the suggestions I am making in the PR is to reduce complexity to an absolute minimum; as it is, this PR introduces 6 entirely new classes and completely refactors 2 existing classes (one of which is absolutely central to the framework). I've no doubt that the code works, I just want to make sure we have achieve it in its most straightforward form 👍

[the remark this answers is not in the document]
jilu1: I forgot to mention that ComposerPackage(r) is for dealing with composer.json which could be installed or not installed and ComposerInstaller is for composer.lock which are installed modules. They also share some seemingly same functions like, to check if a module is a test module, get required and get suggests, check dependencies, etc. But their internal implementation will be different. The former will be parsing the json file, the later will call composer apis. 

Is it a good idea to mingle them into one class?  i.e.
Potential two versions of isMftfTestPackage() in one class?

I know the change is big, but the most complicated part is still in ModuleResolver itself. These 2 classes or 4 classes are minor and I think separate class to deal with composer install package vs composer.json parsed package are more cleaner.

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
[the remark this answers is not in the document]
jilu1: Ok, i will rename it.

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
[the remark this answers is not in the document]
KevinBKozan: It's only a good idea to mingle into one class if we could completely remove one of them, which to your credit is seeming more and more difficult as I dive into the code.

Let me hack away at this and try to see if I can even remove them, I'll need your help for use cases to double check everything still works.

src/Magento/FunctionalTestingFramework/Composer/Objects/ComposerFactory.php
```diff
@@ -0,0 +1,39 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\FunctionalTestingFramework\Composer\Objects;
+
+use Composer\IO\BufferIO;
+
+class ComposerFactory
```
KevinBKozan: I'm not 100% sure we need this class. Only use I found of `create()` is in `AbstractComposerHandler`. We should just call `\Composer\Factory::create(new BufferIO(), $this->composerFile)` directly instead of creating this factory class for a single usage.

src/Magento/FunctionalTestingFramework/Composer/Util/ComposerJsonFinder.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Find absolute paths of all composer json files in a given directory at certain depths
+     *
+     * @param string  $directory
+     * @param integer $depth
+     * @return array
+     */
+    public function finComposerJsonFilesAtDepth($directory, $depth)
```
KevinBKozan: Typo in `fin` => `find`

src/Magento/FunctionalTestingFramework/Composer/Util/ComposerJsonFinder.php
```diff
@@ -0,0 +1,68 @@
...
+ * See COPYING.txt for license details.
+ */
+namespace Magento\FunctionalTestingFramework\Composer\Util;
+
+use Magento\FunctionalTestingFramework\Exceptions\TestFrameworkException;
+
+/**
+ * Class ComposerJsonFinder searches composer json file for possible test module code paths
+ */
+class ComposerJsonFinder
```
KevinBKozan: What do you think about moving both of these functions into `ComposerModuleResolver`? We can abstract out into a utility class when we find the need to use it outside of that context.

src/Magento/FunctionalTestingFramework/Composer/Handlers/AbstractComposerHandler.php
```diff
@@ -0,0 +1,102 @@
...
+     * @var \Composer\Package\CompletePackage
+     */
+    protected $rootPackage;
+
+    /**
+     * @param ComposerFactory $composerFactory
+     */
+    public function __construct(ComposerFactory $composerFactory)
+    {
+        $this->composerFactory = $composerFactory;
```
KevinBKozan: See comment in `ComposerFactory`, I think we can remove this class entirely.

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
```diff
@@ -0,0 +1,130 @@
...
+     * @return array
+     * @throws TestFrameworkException
+     */
+    public function getComposerInstalledTestModules($rootComposerFile)
+    {
+        if (null !== $this->installedTestModules) {
+            return $this->installedTestModules;
+        }
+
+        if (!is_file($rootComposerFile) || substr($rootComposerFile, -13) != 'composer.json') {
```
KevinBKozan: use `basename($rootComposerFile)` instead of `substr`

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
```diff
@@ -0,0 +1,130 @@
...
+        $modules = [];
+        $jsonFinder = new ComposerJsonFinder();
+        $fileList = $jsonFinder->finComposerJsonFilesAtDepth($normalizedDir, 2);
+        foreach ($fileList as $file) {
+            // Parse composer json for test module name and path information
+            $composerInfo = new ComposerPackager(
+                new ComposerFactory($file)
+            );
+            if ($composerInfo->isMftfTestPackage()) {
+                //$moduleName = $composerInfo->getName();
```
KevinBKozan: Remove this line

src/Magento/FunctionalTestingFramework/Util/ModulePathExtractor.php
```diff
@@ -11,41 +11,92 @@
...
+     */
+    private $testModulePaths = [];
+
+    /**
+     * ModulePathExtractor constructor
+     */
+    public function __construct()
+    {
+        $verbose = true;
+        if (empty($this->testModulePaths)) {
```
KevinBKozan: Remove this conditional, `$this->testModulePaths` is never not going to be empty.

src/Magento/FunctionalTestingFramework/Util/ModulePathExtractor.php
```diff
@@ -11,41 +11,92 @@
...
+
+    /**
+     * Extract module name key by path
+     *
+     * @param string $path
+     * @return string
+     */
+    private function extractKeyByPath($path)
+    {
+        if (empty($path)) {
```
KevinBKozan: Function can be simplified:

```
        $shortenedPath = dirname(dirname($path));
        if (empty($path)) {
            return '';
        }
        foreach ($this->testModulePaths as $key => $value) {
            if ($value == $shortenedPath) {
                return $key;
            }
        }
        return '';
```

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -38,6 +39,44 @@ class ModuleResolver
      */
     const REGISTRAR_CLASS = "\Magento\Framework\Component\ComponentRegistrar";
 
+    /**
+     * const for Test/Mftf
```
KevinBKozan: Remove dockblock for `consts`

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -314,46 +429,193 @@ private static function globRelevantWrapper($testPath, $pattern)
...
-                    ['knownDirectory' => $magentoModuleName]
-                );
+        if (null !== $this->composerJsonModulePaths) {
+            return $this->composerJsonModulePaths;
+        }
+        try {
+            $this->composerJsonModulePaths = [];
+            $resolver = new ComposerModuleResolver();
+            $this->composerJsonModulePaths = $resolver->getTestModulesFromPaths($codePaths);
+        } catch (TestFrameworkException $e) {
```
KevinBKozan: Should we at least log these exceptions?

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -314,46 +429,193 @@ private static function globRelevantWrapper($testPath, $pattern)
...
+    private function getComposerInstalledTestModulePaths($composerFile)
+    {
+        if (null !== $this->composerInstalledModulePaths) {
+            return $this->composerInstalledModulePaths;
+        }
+        try {
+            $this->composerInstalledModulePaths = [];
+            $resolver = new ComposerModuleResolver();
+            $this->composerInstalledModulePaths = $resolver->getComposerInstalledTestModules($composerFile);
+        } catch (TestFrameworkException $e) {
```
KevinBKozan: Should we at least log these exceptions?

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -575,4 +839,43 @@ private function getBackendUrl()
...
+    {
+        $possibleVendorName = 'UnknownVendor';
+        $dirPaths = [
+            self::VENDOR,
+            self::APP_CODE,
+            self::DEV_TESTS
+        ];
+
+        foreach ($dirPaths as $dirPath) {
+            $regex = "~.+\\/" . $dirPath . "\/(?<" . self::VENDOR . ">[^\/]+)\/.+~";
```
KevinBKozan: Small refactor for legibility:
```
            $regex = '~/' . $dirPath . '/([^/]+)~';
            // Escape directory separators
            $regex = str_replace('/', '\/', $regex);
            preg_match($regex, $path, $match);
            if (isset($match[1])) {
                $possibleVendorName = ucfirst($match[1]);
                break;
            }
```

What do you think?

src/Magento/FunctionalTestingFramework/Composer/Handlers/AbstractComposerHandler.php
```diff
@@ -0,0 +1,102 @@
...
+namespace Magento\FunctionalTestingFramework\Composer\Handlers;
+
+use Composer\Package\CompletePackageInterface;
+use Magento\FunctionalTestingFramework\ObjectManager;
+use Magento\FunctionalTestingFramework\Composer\Objects\ComposerFactory;
+
+/**
+ *  Abstract Composer Handler
+ */
+abstract class AbstractComposerHandler
```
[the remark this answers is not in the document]
jilu1: Fixed

src/Magento/FunctionalTestingFramework/Composer/Handlers/ComposerInstaller.php
```diff
@@ -0,0 +1,97 @@
...
+ * See COPYING.txt for license details.
+ */
+namespace Magento\FunctionalTestingFramework\Composer\Handlers;
+
+use Composer\Package\CompletePackageInterface;
+
+/**
+ * Class ComposerInstaller handles information and dependencies for composer installed packages
+ */
+class ComposerInstaller extends AbstractComposerHandler
```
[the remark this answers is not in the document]
jilu1: Fixed

src/Magento/FunctionalTestingFramework/Composer/Handlers/ComposerPackager.php
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\FunctionalTestingFramework\Composer\Handlers;
+
+/**
+ * Class ComposerPackage contains composer json information in a MFTF test package
+ */
+class ComposerPackager extends AbstractComposerHandler
```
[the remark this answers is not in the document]
jilu1: Fixed

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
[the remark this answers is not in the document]
jilu1: Renamed

src/Magento/FunctionalTestingFramework/Composer/Objects/ComposerFactory.php
jilu1: Removed the factory class

src/Magento/FunctionalTestingFramework/Composer/Handlers/AbstractComposerHandler.php
jilu1: Done

src/Magento/FunctionalTestingFramework/Composer/Handlers/ComposerPackager.php
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * Get root package
+     *
+     * @return \Composer\Package\RootPackageInterface
+     */
+    public function getRootPackage()
+    {
+        if (!$this->rootPackage) {
```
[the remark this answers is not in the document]
jilu1: Do you mean we can accomplish caching of root package without this rootPackage class variable? If not, considering the number of times the root package is going to be called, I think it's worth to keep it.

src/Magento/FunctionalTestingFramework/Composer/Handlers/AbstractComposerHandler.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @var ComposerFactory
+     */
+    protected $composerFactory;
+
+    /**
+     * @var \Composer\Package\CompletePackage
+     */
+    protected $rootPackage;
```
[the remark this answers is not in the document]
jilu1: Fixed

src/Magento/FunctionalTestingFramework/Composer/Handlers/AbstractComposerHandler.php
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ *  Abstract Composer Handler
+ */
+abstract class AbstractComposerHandler
+{
+    /**
+     * Mftf test module type
+     */
+    const TEST_MODULE_PACKAGE_TYPE = 'magento2-functional-test-module';
```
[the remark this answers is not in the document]
jilu1: Done

src/Magento/FunctionalTestingFramework/Composer/Util/ComposerJsonFinder.php
jilu1: Fixed

src/Magento/FunctionalTestingFramework/Composer/Util/ComposerJsonFinder.php
jilu1: They are actually utils and don't belong to composer or ComposerModuleResolver. but I am ok to move it to ComposerModuleResolver.

src/Magento/FunctionalTestingFramework/Composer/Handlers/AbstractComposerHandler.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Magento module type
+     */
+    const MAGENTO_MODULE_PACKAGE_TYPE = 'magento2-module';
+
+    /**#@+
+     * Composer package array keys
+     */
+    const KEY_PACKAGE_NAME = 'name';
```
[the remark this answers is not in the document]
jilu1: Fixed

src/Magento/FunctionalTestingFramework/Composer/Handlers/ComposerPackager.php
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Parse input array and return all suggested magento module names in pattern like: Magento_Store, Amazon_Core, etc
+     *
+     * @param array $suggests
+     * @return array
+     */
+    private function parseSuggestsForMagentoModuleNames($suggests)
+    {
+        $magentoModuleNames = [];
+        foreach ($suggests as $suggest) {
```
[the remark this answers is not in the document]
jilu1: Your codes as they are probably won't work. As we also need to make sure that it's a "magento2-module" and any white spaces in between. Of course I can build a regex for all of those.

The ultimately question is do we prefer "regex" or "code block"?

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
jilu1: Changed

src/Magento/FunctionalTestingFramework/Util/ComposerModuleResolver.php
jilu1: Done

src/Magento/FunctionalTestingFramework/Util/ModulePathExtractor.php
jilu1: This is intentional. $this->testModulePaths will be queried one time and is cached after

src/Magento/FunctionalTestingFramework/Util/ModulePathExtractor.php
jilu1: Your code will need more checks as well. like what if depth of $path is shorter than 3. let's discuss in person

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
jilu1: Done

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
jilu1: This should be logged by ComposerModuleResolver as an exception automatically, is it sufficient?

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
KevinBKozan: Good point, didn't realize it 👍

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
jilu1: Let's discuss in person

src/Magento/FunctionalTestingFramework/Composer/Handlers/ComposerPackager.php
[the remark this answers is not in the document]
jilu1: Ok, I see the two foreach loops. Let me change this to regex then

src/Magento/FunctionalTestingFramework/Util/ModulePathExtractor.php
KevinBKozan: I don't think so; every place that ModulePathExtractor is used, it's instantiated individually. Since this class isn't a singleton, the instance of the class will always have a nulled out `$this->testModulePaths`.

You can see this by attaching a debugger to line 30; that line is hit multiple times in a single `generate:tests` command.

etc/config/.env.example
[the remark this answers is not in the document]
KevinBKozan: From my reviewing it seems `Magento_Module` format is still usable for `.env` since it merges to an existing list of `Magento_Module` formatted modules.

@jilu1 do either format work based on your implementation? I think the biggest concern here is backward compatibility.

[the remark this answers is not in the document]
jilu1: Before the change, these modules are not found by the ModuleResolver because they did not follow the search pattern.
After change, we search this path specifically. And they are in white list so it will be included.

PR #454, opened 2019-09-16
MQE-1765: Introduce API Endpoint and Request Headers to Allure artifacts

KevinBKozan:
### Description
- Removed both artifacts to be reintroduced after bugfixes for duped filename

### Contribution checklist


src/Magento/FunctionalTestingFramework/DataGenerator/Persist/CurlHandler.php
```diff
@@ -167,8 +167,6 @@ public function executeRequest($dependentEntities)
         $response = $executor->read($successRegex, $returnRegex, $returnIndex);
         $executor->close();
 
-        AllureHelper::addAttachmentToLastStep($apiUrl, 'API Endpoint');
```
tomreece: Why do we only strip out part of it?

KevinBKozan: The problem is very specific to the way php-allure comes up with filenames. It uses a SHA based on the contents passed in, so unique contents are entirely unaffected (body and data will almost certainly be unique)

Allure really doesn't like it when it references the same filename in two different steps. It doesn't break the report, but it does cause some very weird behavior.

tomreece: Gotcha! 👍

PR #460, opened 2019-09-19
MQE-1763: Investigate and fix failure due to absence of AcceptanceTester class

tomreece:
### Description
This bug was reproducible on branches that had updated the `symfony/yaml` package in composer from v3.4.30 to v4.3.4. The updated package was less forgiving about our functional.suite.yml file which did not properly quote string values in some cases.

This PR only exists to ensure that in the future we get better output from the `mftf build:project` command. This output was hidden in the Jenkins job because the `--verbose` flag was not used. If we had that output in Jenkins, then this bug would have been trivial to identify.

I also added troubleshooting steps to devdocs for users on versions lower than MFTF 2.5.0.

TLDR: No bug here. I increased output from `mftf build:project` and wrote some docs.

1. https://jira.corp.magento.com/browse/MQE-1763

### Contribution checklist

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -61,39 +61,5 @@ Use of PhantomJS is not supported by the MFTF.
 
 For headless browsing, the [Headless Chrome][]{:target="\_blank"} has better compatibility with the MFTF.
 
-### Chrome
-
-You are seeing an "unhandled inspector error" exception:
-
-```terminal
-[Facebook\WebDriver\Exception\UnknownServerException]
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -2,60 +2,64 @@
 
 Having a little trouble with the MFTF? See some common errors and fixes below.
 
-## WebDriver issues
```
tomreece: This documentation change looks more complicated than it is.

I discussed with @dobooth and we agreed to remove the two existing troubleshooting items on this page because they were no longer relevant.

PR #489, opened 2019-10-17
MQE-1644: Add ability to see JS log in Allure 

KevinBKozan:
### Description
- Refactored ErrorLogger (no longer singleton, renamed for its actual use
- Added `stepAfter` actions to log browser logs

### Contribution checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -277,6 +277,24 @@ Example:
 CREDENTIAL_VAULT_SECRET_BASE_PATH=secret
 ```
 
+### ENABLE_BROWSER_LOG
+
+Enables addition of browser logs to Allure steps
+
+```conf
+ENABLE_BROWSER_LOG=true
+```
+
+### BROWSER_LOG_BLACKLIST
+
+Blacklists types of browser log entries from appearing in Allure steps.
+
+Denoted in browser log entry as `"SOURCE": "type"`.
+
+```conf
+BROWSER_LOG_BLACKLIST=other,console-api
+```
+
 <!-- Link definitions -->
 
 [`MAGENTO_CLI_COMMAND_PATH`]: #magento_cli_command_path
--- a/etc/config/.env.example
+++ b/etc/config/.env.example
@@ -54,4 +54,9 @@ MODULE_WHITELIST=Magento_Framework,ConfigurableProductWishlist,ConfigurableProdu
 
 #*** Default timeout for wait actions
 #WAIT_TIMEOUT=10
+
+#*** Uncomment and set to enable browser log entries on actions in Allure. Blacklist is used to filter logs of a specific "source"
+#ENABLE_BROWSER_LOG=true
+#BROWSER_LOG_BLACKLIST=other
+
 #*** End of .env ***#
--- a/src/Magento/FunctionalTestingFramework/Extension/BrowserLogUtil.php
+++ b/src/Magento/FunctionalTestingFramework/Extension/BrowserLogUtil.php
@@ -16,23 +16,20 @@ class BrowserLogUtil
     const ERROR_TYPE_JAVASCRIPT = "javascript";
 
     /**
-     * Loops through stepEvent for browser log entries
+     * Loops throw errors in log and logs them to allure. Uses Module to set the error itself
      *
-     * @param \Magento\FunctionalTestingFramework\Module\MagentoWebDriver $module
-     * @param \Codeception\Event\StepEvent                                $stepEvent
+     * @param array                         $log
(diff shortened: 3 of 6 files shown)


docs/configuration.md
```diff
@@ -277,6 +277,24 @@ Example:
...
+
+Enables addition of browser logs to Allure steps
+
+```conf
+ENABLE_BROWSER_LOG=true
+```
+
+### BROWSER_LOG_BLACKLIST
+
+Blacklists types of browser log entries from appearing in Allure steps.
```
jilu1: Can we add all possible types in the blacklist? are `other` and `console-api` the only possible types?

etc/config/.env.example
```diff
@@ -54,4 +54,9 @@ MODULE_WHITELIST=Magento_Framework,ConfigurableProductWishlist,ConfigurableProdu
 
 #*** Default timeout for wait actions
 #WAIT_TIMEOUT=10
+
+#*** Uncomment and set to enable browser log entries on actions in Allure. Blacklist is used to filter logs of a specific "source"
+#ENABLE_BROWSER_LOG=true
+#BROWSER_LOG_BLACKLIST=other
```
jilu1: `console-api` is not in the blacklist, is this on purpose?

src/Magento/FunctionalTestingFramework/Extension/BrowserLogUtil.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Loops through given log and filters entries of the given type.
+     *
+     * @param array  $log
+     * @param string $type
+     * @return array
+     */
+    public static function filterLogsOfType($log, $type)
```
jilu1: `filterLogsOfType()` and `getLogsOfType` are complete the same. Any reason we need two functions here?

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
```diff
@@ -173,7 +174,17 @@ public function beforeStep(\Codeception\Event\StepEvent $e)
      */
     public function afterStep(\Codeception\Event\StepEvent $e)
     {
-        ErrorLogger::getInstance()->logErrors($this->getDriver(), $e);
+        if (getenv('ENABLE_BROWSER_LOG')) {
+            $browserLog = $this->getDriver()->webDriver->manage()->getLog("browser");
+            foreach (explode(',', getenv('BROWSER_LOG_BLACKLIST')) as $source) {
+                $browserLog = BrowserLogUtil::filterLogsOfType($browserLog, $source);
+            }
```
jilu1: Can line 177 ~ 181 be moved to BrowserLogUtil class? The same logic is in BrowserLogUtil.

docs/configuration.md
KevinBKozan: I'm not 100% sure what all of the types are, I've ony been able to trigger those two.

If you can find documentation of the available types then please let me know, I've had no luck.

etc/config/.env.example
KevinBKozan: Kind of, yes. `other` maps to miscelaneous Chrome specific warnings as far as I can tell, which wouldn't be particularly useful.

Do you think thi sblacklist should be blak by default?

src/Magento/FunctionalTestingFramework/Extension/BrowserLogUtil.php
KevinBKozan: They are almost completely the same, check ln50 vs ln68

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
KevinBKozan: Didn't watch this, I'll see if I can move it

KevinBKozan: I did it the other way around, I removed redundant logic from `BrowserLogUtil::logErrors()`. This kind of log getting should be done by the consumer of the Util, not by the util itself in my opinion.

src/Magento/FunctionalTestingFramework/Extension/BrowserLogUtil.php
jilu1: Ok, I see the difference.

docs/configuration.md
jilu1: I see. Then let's not mention.

etc/config/.env.example
jilu1: If it's on purpose, I am fine. No need to change.

PR #507, opened 2019-11-14
MQE-1884: MFTF - <after> failures override other failures

soumyau:
attaching suppressed exception before _after hook failure to current step.

2. ...

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
+++ b/src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
@@ -36,7 +36,6 @@ class TestContextExtension extends BaseExtension
     {
         $events = [
             Events::TEST_START => 'testStart',
-            Events::TEST_FAIL => 'testFail',
             Events::STEP_AFTER => 'afterStep',
             Events::TEST_END => 'testEnd',
             Events::RESULT_PRINT_AFTER => 'saveFailed'
@@ -57,18 +56,7 @@ class TestContextExtension extends BaseExtension
     }
 
     /**
-     * Codeception event listener function, triggered on test failure.
-     * @param \Codeception\Event\FailEvent $e
-     * @return void
-     */
-    public function testFail(\Codeception\Event\FailEvent $e)
-    {
-        //log suppressed exception in case of _after hook failure
-        $this->logPreviousException($e->getFail());
-    }
-
-    /**
-     * Codeception event listener function, triggered on test ending (naturally or by error).
+     * Codeception event listener function, triggered on test ending naturally or by errors/failures.
      * @param \Codeception\Event\TestEvent $e
      * @return void
      * @throws \Exception
@@ -77,20 +65,28 @@ class TestContextExtension extends BaseExtension
     {
         $cest = $e->getTest();
 
-        //Access private TestResultObject to find stack and if there are any errors (as opposed to failures)
+        //Access private TestResultObject to find stack and if there are any errors/failures
         $testResultObject = call_user_func(\Closure::bind(
             function () use ($cest) {
                 return $cest->getTestResultObject();
             },
             $cest
         ));
-        $errors = $testResultObject->errors();
-        if (!empty($errors)) {
-            foreach ($errors as $error) {
-                if ($error->failedTest()->getTestMethod() == $cest->getName()) {
-                    //log suppressed exception in case of _after hook failure
-                    $this->logPreviousException($error->thrownException());
-                    continue;
+
+        // check for errors in all test hooks and attach in allure
+        if (!empty($testResultObject->errors())) {
+            foreach ($testResultObject->errors() as $error) {
+                if($error->failedTest()->getTestMethod() == $cest->getTestMethod()) {
+                    $this->attachExceptionToAllure($error->thrownException(), $cest->getTestMethod());
+                }
+            }
+        }
+
+        // check for failures in all test hooks and attach in allure
+        if (!empty($testResultObject->failures())) {
+            foreach ($testResultObject->failures() as $failure) {
+                if($failure->failedTest()->getTestMethod() == $cest->getTestMethod()) {
+                    $this->attachExceptionToAllure($failure->thrownException(), $cest->getTestMethod());
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
```diff
@@ -150,6 +155,30 @@ public function extractContext($trace, $class)
...
+    {
+        $change = function () {
+            if ($this instanceof \PHPUnit\Framework\ExceptionWrapper) {
+                return $this->previous;
+            } else {
+                return $this->getPrevious();
+            }
+        };
+        $firstException = $change->call($exception);
+        $bind = function () use ($firstException) {
```
KevinBKozan: I'm not sure what 173~176 do. `ExceptionWrapper.php` does not have a variable named `$exception`, and at this point you already have the thing you need to log.

Is this leftover test flow or something like that?

KevinBKozan:
On further experimentation, I am not sure I am seeing desired behavior with Codeception 2.4.0+:
image

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
soumyau: removed the redundant code.

soumyau:
> Left some feedback purely on the implementation side.
> 
> I think we need a little more time to test this. In the case of two comparison/errors I think we are good to go:
> image
> 
> In the case of two broken steps, I do not think so, only the first broken step is injected into the test.



> On further experimentation, I am not sure I am seeing desired behavior with Codeception 2.4.0+:
> image
Standalone mftf (codeception 2.3.9) does not seem to report broken steps correctly. After updating mftf with codeception 2.4.5, standalone mftf reports:
MQE-1884-5
MQE-1884-4
MQE-1884-3
MQE-1884-2
MQE-1884-1

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
```diff
@@ -150,6 +115,26 @@ public function extractContext($trace, $class)
         return null;
     }
 
+    /**
+     * Attach suppressed exception thrown before _after hook to the current step.
+     * @param  \Exception $exception
+     * @return mixed
+     */
+    public function logPreviousException(\Exception $exception)
```
KevinBKozan: Question:
* Should we attach every exception, *not* just the buried one? It feels strange to see a test fail in two steps but only see one `Exception` attached. We could even label it `Before/Test/After Exception` using some functions in this class already.

soumyau: Those were my initial thoughts too, but looks like `After` block's exception stack trace appears in the allure report twice - one at the top and another at the bottom after test steps. Will it be a duplication in some way? Capturing context and appending it to screenshot would work, however, there will only be 1 exception in `Before + Test body` scope, would context  be redundant to display? Also looks like Allure doesn't a clear separation of `Before/Test/After` blocks  of the test. What are your thoughts?

KevinBKozan: So good news is that with just Codeception 2.4.0 we can do a couple things:
* Stop iterating over errors on line 88, they will only now be parent->previous() relationship
* Given above, we can instead iterate by doing logError($error), wherein we log the parent, check for child and call logError($childError) to pop the full stack
* remove the `testFail()` hook entirely, just use `testEnd()` to log these

We can modify the hook to look something like:

```
...
$testResultObject = call_user_func(\Closure::bind(
    function () use ($cest) {
        return $cest->getTestResultObject();
    },
    $cest
));
if (!empty($testResultObject->errors())) {
    foreach ($testResultObject->errors() as $error) {
        $this->attachExceptionToAllure($error->thrownException());
    }
}
if (!empty($testResultObject->failures())) {
    foreach ($testResultObject->failures() as $error) {
        $this->attachExceptionToAllure($error->thrownException());
    }
}
...
```

We can end up with something like this
image

You could add to the signature of `attachExceptionToAllure` to inlcude a prefix like `Error` vs `Failure` as well for clearer deliniation.

KevinBKozan: Even if we have some duplication, I think if we lean on prefixing the attachments it would be best in case someone needs the info. Those can be ignored by simply not expanding them.

soumyau: Added context as a prefix to the attachment. Including a prefix like `Error` or `Failure` may need more investigation. From what I understood, `PHPUnit\Framework\Exception` are marked as failures and `Magento\Framework\Exception` are marked as Errors. Maybe there are exceptions which don't follow this criteria.
Screen Shot 2019-11-19 at 9 22 53 AM

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
```diff
@@ -116,22 +112,36 @@ public function extractContext($trace, $class)
...
-    public function logPreviousException(\Exception $exception)
+    public function attachExceptionToAllure($exception, $testMethod)
     {
+        if (is_subclass_of($exception, \PHPUnit\Framework\Exception::class)) {
+            $trace = $exception->getSerializableTrace();
+        } else {
+            $trace = $exception->getTrace();
+        }
+
+        $context = $this->extractContext($trace, $testMethod);
```
KevinBKozan: How do you feel about doing a mapping here?
```
const TEST_HOOKS = ['_after' => 'AfterHook']
...
if (isset(self::TEST_HOOKS[$context])) {
    $context = self::TEST_HOOKS[$context];
} else {
    $context = 'Test'
}
```

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
KevinBKozan: I think you are right in not trusting that the exception types are necessarily linked to failure vs error. You could include them in the signature of your new `attachExceptionToAllure` but I think having the prefix is enough for now.

src/Magento/FunctionalTestingFramework/Extension/TestContextExtension.php
soumyau: Added mappings, modified 'Test' to 'TestMethod' as per Cest.php verbiage. Open to changing it back to 'Test' if you think it'll be better.
image

PR #509, opened 2019-11-20
MQE-1470: add interface to handle file path and url format

jilu1:

2. ...

### Contribution checklist


src/Magento/FunctionalTestingFramework/Util/Path/FileUrlFormatterInterface.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\FunctionalTestingFramework\Util\Path;
+
+use Magento\FunctionalTestingFramework\Exceptions\TestFrameworkException;
+
+interface FileUrlFormatterInterface
```
KevinBKozan: Change this to `FormatterInterface`

src/Magento/FunctionalTestingFramework/Util/Path/UrlFormatter.php
```diff
@@ -0,0 +1,103 @@
...
+     * @param boolean $withTrailingSeparator
+     * @return string
+     * @throws TestFrameworkException
+     */
+    public static function format($url, $withTrailingSeparator = true)
+    {
+        $sanitizedUrl = rtrim($url, '/');
+        $sanitizedUrl = filter_var($sanitizedUrl, FILTER_SANITIZE_URL);
+
+        if (false === $sanitizedUrl) {
```
KevinBKozan: Please add some comments on each of the lines before the `if` statements (where applicable)
Something like `Check if URL is in format: <format>`

src/Magento/FunctionalTestingFramework/Util/Path/FilePathFormatter.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Return formatted full file path from input string, or false on error
+     *
+     * @param string  $path
+     * @param boolean $withTrailingSeparator
+     * @return string
+     * @throws TestFrameworkException
+     */
+    public static function format($path, $withTrailingSeparator = false)
```
KevinBKozan: I do not think that this function should have an option for trailing separator. In my eyes, the point of this is that it will *always* return a standardized file path, so *always* with/without trailing separator.

KevinBKozan: Same applies to URL (didn't see it at first)

src/Magento/FunctionalTestingFramework/Util/Path/FileUrlFormatterInterface.php
jilu1: Changed

src/Magento/FunctionalTestingFramework/Util/Path/UrlFormatter.php
jilu1: Added

src/Magento/FunctionalTestingFramework/Util/Path/FilePathFormatter.php
jilu1: Personally I prefer this as a optional parameter with default to true. Further code refactoring shows that about 90% of time we need the /. With this as an option gives the flexibility to choose and eliminates the use of trim() in calling function.

src/Magento/FunctionalTestingFramework/Util/Path/FileUrlFormatterInterface.php
tomreece: What benefit do we get from having this interface with only one function?

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
```diff
@@ -201,7 +202,7 @@ public function getEnabledModules()
 
         $token = $this->getAdminToken();
 
-        $url = ConfigSanitizerUtil::sanitizeUrl(getenv('MAGENTO_BASE_URL')) . $this->moduleUrl;
+        $url = UrlFormatter::format(getenv('MAGENTO_BASE_URL')) . $this->moduleUrl;
```
tomreece: Instead of a generic format() function, I expected the new class to operate like:

`AwesomeNewClass::getMagentoBaseUrl()`
`AwesomeNewClass::getBackendUrl()`

Maybe I was thinking wrong? What do you think?

src/Magento/FunctionalTestingFramework/Util/Path/FilePathFormatter.php
tomreece: I don't mind the argument as long as it defaults to the most common use case.

Right now it defaults to false but @jilu1 says she wants it to default to true. So will you be changing it to true @jilu1 ?

src/Magento/FunctionalTestingFramework/Util/Path/FileUrlFormatterInterface.php
jilu1: To me this interface is just to group two type of formatter class. They have similarity but unrelated at the same time. What's the rule for interface?

src/Magento/FunctionalTestingFramework/Util/Path/FilePathFormatter.php
jilu1: @tomreece Yes. see updated version

src/Magento/FunctionalTestingFramework/Util/Path/FileUrlFormatterInterface.php
tomreece: I'm no OOP expert. I just don't like the existence of useless things to exist.

I see the value of an interface being when you have multiple implementations and multiple functions within.

 I just don't like it but probably someone else is best to make the final call on this.

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
jilu1: In the refactored code, you can see that we have a few places that you can query magentoBaseUrl() or backendUrl(). 
This interface util is to standardize the use of user specified variable regardless of trailing /, etc. If you look at the various refactored code, you can see the vulnerability of the way we use user specified envs, e.g. TESTS_MODULE_PATH, TESTS_BP, etc

src/Magento/FunctionalTestingFramework/Util/Path/FilePathFormatter.php
KevinBKozan: Cool, I'm fine with the argument if it's set to default usecase (which you both agree on).

src/Magento/FunctionalTestingFramework/Util/ModuleResolver.php
tomreece: Gotcha.

PR #512, opened 2019-11-21
MQE-1520: Consolidate WAIT_TIMEOUT in .env and pageload_timeout

KevinBKozan:
### Description
* Added WAIT_TIMEOUT to default yml file + default definition as this makes it required to be defined
* Removed some dead code since it's always defined now.


dev/tests/_bootstrap.php
```diff
@@ -45,7 +45,8 @@
     'MAGENTO_BACKEND_NAME' => 'admin',
     'MAGENTO_ADMIN_USERNAME' => 'admin',
     'MAGENTO_ADMIN_PASSWORD' => 'admin123',
-    'DEFAULT_TIMEZONE' => 'America/Los_Angeles'
+    'DEFAULT_TIMEZONE' => 'America/Los_Angeles',
+    'WAIT_TIMEOUT' => '10'
```
jilu1: Do we want this WAIT_TIMEOUT to be defined in .env ? or hiding from user in bootstrap?

dev/tests/functional/standalone_bootstrap.php
```diff
@@ -49,7 +49,10 @@
 
     defined('DEFAULT_TIMEZONE') || define('DEFAULT_TIMEZONE', 'America/Los_Angeles');
     $env->setEnvironmentVariable('DEFAULT_TIMEZONE', DEFAULT_TIMEZONE);
-    
+
+    defined('WAIT_TIMEOUT') || define('WAIT_TIMEOUT', 10);
+    $env->setEnvironmentVariable('WAIT_TIMEOUT', 10);
```
jilu1: Is 10 seconds sufficient for page load? especially when page is not cached.

dev/tests/_bootstrap.php
jilu1: Never mind. WAIT_TIMEOUT is in .env.example.

jilu1: Let's add WAIT_TIMEOUT in Dev Doc.

KevinBKozan: Yup, I'll add it to devdocs!

dev/tests/functional/standalone_bootstrap.php
KevinBKozan: Honestly I don't think so. We can change this default to 30.
I changed it to 10 as that's what it was in ActionObject, that may have been the wrong call.

tomreece:
@jilu1 Are you sure that WAIT_TIMEOUT is required now?
it looks to me like defaults are provided but I haven’t ran it locally to confirm.

If it is in fact required, I think we should undo that and make it NOT required. We should provide a sensible default of 30 seconds or something like that.

jilu1:
I saw it from a different perspective that WAIT_TIMEOUT is always required now but it’s Ok with me if we want to "hide" it from .env.

PR #516, opened 2019-11-27
MQE-1379: Fix MFTF custom actions to fully support Codeception dry-ru…

jilu1:
…n functionality

2. ...

### Contribution checklist


src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
```diff
@@ -0,0 +1,157 @@
...
+ * @package Magento\FunctionalTestingFramework\Module
+ */
+class MagentoActionProxies extends CodeceptionModule
+{
+    /**
+     * PersistedObjectHandler instance
+     *
+     * @var PersistedObjectHandler
+     */
+    private static $persistHandler = null;
```
KevinBKozan: Storing a reference to a singleton does not play into the advantages of the design pattern; the whole point of it is that any time we want to reference the singleton we can `Singleton::getInstance()` without instantiating it (or having to keep a reference to an instantiated singleton.

I would strongly reccomend removal of this private static variable, as well as removing all of the `if (!self::persistHandler)` clauses. This should work to keep this file even more trim than it already is.

I have the feeling I'm missing some partial context here, so please feel free to enlighten me if I'm missing a key reason for this design decision.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Retrieves a field from an entity, according to key and scope given
+     *
+     * @param string $stepKey
+     * @param string $field
+     * @param string $scope
+     * @return string
+     */
+    public function retrieveEntityField($stepKey, $field, $scope)
```
KevinBKozan: I'm not sure I understand the reason for this wrapper function, does `dry-run` throw errors if it encounters `PersistedObjectHandler::getInstance()->retrieveEntityField()` outside an acceptance tester action?

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Get encrypted value by key
+     *
+     * @param string $key
+     * @return string|null
+     * @throws TestFrameworkException
+     */
+    public function getSecret($key)
```
KevinBKozan: See above comment.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
jilu1: I am aware of the singleton nature of PersistedObjectHandler, and was trying to make it more explicit. I guess it's not necessary. :)

Removed.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
jilu1: This is needed when test trying to referencing the persisted variable. It's not really persisted and won't find it.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
jilu1: Same reason above.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Class MagentoActionProxies
+ *
+ * Contains all proxy functions whose corresponding MFTF actions need to be accessible for AcceptanceTester $I
+ *
+ * @package Magento\FunctionalTestingFramework\Module
+ */
+class MagentoActionProxies extends CodeceptionModule
+{
+    /**
```
KevinBKozan: We should just remove the code instead of commenting it out, and include a `//TODO` line to add the code back in.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
KevinBKozan: image
Changing this to calls to `$I->` has an interesting side-effect (see above)
This is a test that has one `createData` and one `comment input="$createdData.Key$"`

If we absolutely *have* to include this `$I->` wrapper for any `retrieveEntityField`, we need to add an `if(step action = "retrieveEntityField")` clause to MagentoAllureAdapter so that it doesn't print out each retrieval as its own step.

This *also* has to be done with `getSecret`

src/Magento/FunctionalTestingFramework/Util/TestGenerator.php
```diff
@@ -759,7 +765,7 @@ public function generateStepsPhp($actionObjects, $generationScope = TestGenerato
                         $scope = PersistedObjectHandler::SUITE_SCOPE;
                     }
 
-                    $createEntityFunctionCall = "\t\tPersistedObjectHandler::getInstance()->createEntity(";
+                    $createEntityFunctionCall = "\t\t\${$actor}->createEntity(";
```
KevinBKozan: For Allure cleanliness and since entity creation is now an `$I` function, we should do the following:
* Make the `createEntity` line be a one line long (to facilitate the next point)
* Print out stepKey at the end of the line like all other actions (for MagentoAllureAdapter to pick up)
* Remove the inclusion of a comment line before creating the Entity (see line 736)

This should also be done for all of the entity CRUD operations for cleanliness.

KevinBKozan:
@jilu1 One more note, the `generateDate` action does not show up in dry-run (it's a similar case of not being a native `$I` action. It should be easy to fix this.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
jilu1: Ah, good catch! let me looking into this.

jilu1: Removed "retrieveEntityField" and "getSecret" from step display.

src/Magento/FunctionalTestingFramework/Util/TestGenerator.php
jilu1: Done

jilu1:
"generateDate" will generate built in php functions. It won't break the dry run. I think it will behave the same as Codeception as it's now.

src/Magento/FunctionalTestingFramework/Module/MagentoActionProxies.php
jilu1: Done

etc/config/functional.suite.dist.yml
```diff
@@ -16,6 +16,7 @@ modules:
         - \Magento\FunctionalTestingFramework\Helper\MagentoFakerData
         - \Magento\FunctionalTestingFramework\Module\MagentoSequence
         - \Magento\FunctionalTestingFramework\Module\MagentoAssert
+        - \Magento\FunctionalTestingFramework\Module\MagentoActionProxies
```
tomreece: @jilu1 Does this change affect the pipeline code since they have their own version of `functional.suite.yml`?

tomreece: From Ji in slack:

> Currently the functional.suite.dist.yml change is not used. In a later story, we will remove this dependency. I think I have a ticket number in the dry-run ticket for that work.

PR #538, opened 2020-01-06
FEATURE: <magentoCron> command to execute Cron Jobs

lbajsarowicz:
### Description
Command executes Cron Jobs taking into consideration need of having interval between running same groups (60 seconds).

Untitled_ Jan 6 2020 3_56 PM - Edited

Use case:
```xml

<tests xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/testSchema.xsd">
    <test name="CronjobsIntervalTest">
        <magentoCron stepKey="runCronjobs1" groups="index stage"/>
        <magentoCron stepKey="notOverlappingCron" groups="magic"/>
        <comment userInput="Some Other Actions" stepKey="otherActions"/>
        <wait stepKey="otherActionsTime" time="10"/>
        <magentoCron stepKey="runCronjobs2"/>
        <magentoCron stepKey="runCronjobs3" groups="index stage"/>
    </test>
</tests>

```

N/A

### Contribution checklist


lbajsarowicz:
What I want to achieve is to avoid constructs like in this file: https://github.com/magento/magento2/blob/0262f1428561f5c329a50bfbd62714fa58e175be/app/code/Magento/Catalog/Test/Mftf/Test/AdminProductCategoryIndexerInUpdateOnScheduleModeTest.xml

Also want to reduce amount of work necessary to work it out:
image

lbajsarowicz:
My first approach used ActionGroup, but then @VladimirZaets asked me to cover my changes also for Magento EE / Magento B2B. I realised that you've introduced running Cron Jobs in... Action Groups:

image

According to the fact, that you cannot have nested ActionGroups, I had to find another approach that would let me execute Cron Jobs having in mind that should not be ran more often than with 1 minute interval.

@okolesnyk Asked me "What if Magento change the way that `bin/magento cron:run` works? MFTF version will be coupled to Magento version.

So my answer is: Not, actually not. If Magento will change the way that `cron:run` works - with current approach, you'll be forced to update hundreds of Tests and Action Groups. Currently - using YAGNI approach, I would'nt bother with this case. But then - when Magento would change the behavior of `cron:run` I'd just gather the Magento version along with getting API Keys and I'd choose the strategy to work with `<magentoCron group="x y z">` to be compatible with the version.

lbajsarowicz:
What also comes to my mind is the fact that not always Cron Jobs end up correctly - sometimes it ends with "Exceeded memory limit" or other errors. With `magentoCron` we can make assertions on command output to make sure that no unnecessary output was thrown from the application.

lbajsarowicz:
Sure. Will handle that tonight.

lbajsarowicz:
@soumyau Done.

PR #542, opened 2020-01-09
MQE-1317: Generate/run test in single suite context

KevinBKozan:
### Description
* Added support for `suite:test` syntax

### Contribution checklist


src/Magento/FunctionalTestingFramework/Suite/SuiteGenerator.php
```diff
@@ -178,7 +178,7 @@ private function validateTestsReferencedInSuite($suiteName, $testsReferenced, $o
     {
         $suiteRef = $originalSuiteName ?? $suiteName;
         $possibleTestRef = SuiteObjectHandler::getInstance()->getObject($suiteRef)->getTests();
-        $errorMsg = "Cannot reference tests whcih are not declared as part of suite.";
+        $errorMsg = "Cannot reference tests which are not declared as part of suite.";
```
jilu1: Can we display $testsReferenced in the error message?

dev/tests/unit/Magento/FunctionalTestFramework/Console/BaseGenerateCommandTest.php
```diff
@@ -156,6 +156,24 @@ public function testThreeTestOneSuiteOneGroupMix()
...
+        );
+
+        $testArray = ['Test1' => $testOne];
+        $suiteArray = ['Suite1' => $suiteOne];
+        $this->mockHandlers($testArray, $suiteArray);
+        $actual = json_decode($this->callTestConfig(['Suite1:Test1']), true);
+        $expected = ['tests' => null, 'suites' => ['Suite1' => ['Test1']]];
+        $this->assertEquals($expected, $actual);
+    }
+
```
jilu1: Can we add negative tests for "Suite ${objectName} is not defined in xml" and "Cannot reference tests which are not declared as part of suite."?

src/Magento/FunctionalTestingFramework/Suite/SuiteGenerator.php
KevinBKozan: I believe so, I'll try to add this.

dev/tests/unit/Magento/FunctionalTestFramework/Console/BaseGenerateCommandTest.php
KevinBKozan: I'll give it a shot. I'm not 100% sure where that exception comes from (it may come from the manifest creation classes, which may mean a separate set of unit tests)

KevinBKozan: Fixed

src/Magento/FunctionalTestingFramework/Suite/SuiteGenerator.php
KevinBKozan: Fixed

dev/tests/unit/Magento/FunctionalTestFramework/Suite/SuiteGeneratorTest.php
```diff
@@ -143,6 +145,49 @@ public function testGenerateEmptySuite()
...
+
+        // Set up Expected Exception
+        $this->expectException(TestReferenceException::class);
+        $this->expectExceptionMessageRegExp('(Suite: "Suite2" Tests: "Test1")');
+
+        // parse and generate suite object with mocked data and manifest
+        $mockSuiteGenerator = SuiteGenerator::getInstance();
+        $mockSuiteGenerator->generateAllSuites($manifest);
+    }
+
```
jilu1: This test is good for the 1st exception case. To test the 2nd exception case, you can use a non-existing suite name and you don't even need to mock any suite/test structure.

PR #551, opened 2020-01-14
#551 MC-30228: [MFTF] Need to specify default store code for createEntity

zakdma:
### Bug - P2
* [MC-30228](https://jira.corp.magento.com/browse/MC-30228) [MFTF] Need to specify default store code for createEntity

### Tests
https://jira.corp.magento.com/browse/MC-30228?focusedCommentId=1350646&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1350646

### Contribution checklist

Created by : @svitja
Internal PR: https://github.com/magento-tsg/magento2-functional-testing-framework/pull/1


zakdma:
@KevinBKozan please review PR

KevinBKozan:
@zakdma Thanks for the contribution 👍 I'll review and keep you updated on testing progress.

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/PersistedObjectHandler.php
```diff
@@ -80,7 +80,7 @@ public function createEntity(
         $entity,
         $dependentObjectKeys = [],
         $overrideFields = [],
-        $storeCode = ""
+        $storeCode = "all"
```
jilu1: @zakdma If I remember correctly, not all API support "all" store code. The default should be "default" or leave empty. Have you encountered any problem with no store code specified?

ameysar: @jilu1 please refer to [app/code/Magento/Webapi/Controller/PathProcessor.php](https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/Webapi/Controller/PathProcessor.php).
**All** API requests support it. It is used to process both *[REST](https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/Webapi/Controller/Rest.php#L182)* and *[SOAP](https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/Webapi/Controller/Soap.php#L124)* API requests.
If we leave store code empty - it will take "current store" ( **default** ) from store manager, but we need to save entity in global scope,
the same as if it was saved in admin with **Default Config** scope.

jilu1: @ameysar I am not convinced that persisting to all stores is desired behavior. Can you just set `storeCode` attribute for your test instead.

ameysar: @jilu1 according to [REST API endpoints](https://devdocs.magento.com/guides/v2.3/get-started/gs-web-api-request.html#endpoints) store code **all** is applied to CMS and Product modules. Is there a possibility to set default store code for **create** operation ( and do not specify it each time for **createData** action in test ) - as it was done in [MTF](https://github.com/magento/magento2ce/blob/2.3-develop/dev/tests/functional/tests/app/Magento/Catalog/Test/Handler/CatalogProductSimple/Webapi.php#L106)?

jilu1: @ameysar Even for create products, creating to **all** stores may not be the desire outcome that everyone wants. I think it makes sense for you to set the store code for your own test. Do you agree?

ameysar: @jilu1 No, I don't agree. It works well in [MTF](https://github.com/magento/magento2ce/blob/2.3-develop/dev/tests/functional/tests/app/Magento/Catalog/Test/Handler/CatalogProductSimple/Webapi.php#L106). Also when you save a product in admin panel - **Default Config** scope is selected by default, so **ALL** products are saved with **admin** store code.

jilu1: @ameysar I will discuss this with my team. I will update you with our decision.

jilu1: @ameysar Our team thinks keeping default value as "default" fits most of the use case and we will not accept your pull request.

PR #560, opened 2020-01-25
MFTF deprecation notice attributes

okolesnyk:

2. ...

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Page/Handlers/SectionObjectHandler.php
+++ b/src/Magento/FunctionalTestingFramework/Page/Handlers/SectionObjectHandler.php
@@ -100,7 +100,7 @@ class SectionObjectHandler implements ObjectHandlerInterface
             $sectionDeprecated = $sectionData[self::OBJ_DEPRECATED] ?? null;
 
             if ($sectionDeprecated !== null) {
-                LoggingUtil::getInstance()->getLogger(ElementObject::class)->deprecation(
+                LoggingUtil::getInstance()->getLogger(SectionObject::class)->deprecation(
                     $sectionDeprecated,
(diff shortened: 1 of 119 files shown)


KevinBKozan:
@okolesnyk Feature is still printing out *all* deprecated entities on generation; we need to either:
* Suppress printing selectively for our new deprecation notices (done by changing function in `MftfLoggingUtil`)
* Detect existance of deprecated entities *once* and only print out a single `Deprecated entities detected in MFTF Test files, please check <log file>`. This is trickier, would need to keep track of if we've thrown that error at all.

Option one is easier, and works to reduce noise. I think that once we implement this, the deprecated warning will literally always be up once it starts being used, so I actually think we can supress *all* deprecated test file warnings.

src/Magento/FunctionalTestingFramework/Page/Handlers/SectionObjectHandler.php
```diff
@@ -73,22 +75,43 @@ private function __construct()
...
             } catch (XmlException $exception) {
                 throw new XmlException($exception->getMessage() . " in Section '{$sectionName}'");
             }
 
             $filename = $sectionData[self::FILENAME] ?? null;
-            $this->sectionObjects[$sectionName] = new SectionObject($sectionName, $elements, $filename);
+            $sectionDeprecated = $sectionData[self::OBJ_DEPRECATED] ?? null;
+
+            if ($sectionDeprecated !== null) {
+                LoggingUtil::getInstance()->getLogger(ElementObject::class)->deprecation(
```
KevinBKozan: Missed on first review, should be `SectionObject::class` I believe.

PR #576, opened 2020-02-05
Removes reference files from MFTF repo 

dobooth:
### Description

This PR removes two files from the repo. They will be hosted in the devdocs repo instead to reduce complexity.
Check with Don before merging.

### Contribution checklist


dobooth:
Couple reasons: 1) These are really codebase references, rather than MFTF references, so devdocs is a better place for it. 2) The topics are in the MFTF repo but the data file and build logic is on devdocs. This change will keep all the files in devdocs and make the process more stable. There was a Git issue with the MFTF repo a couple weeks ago that broke the devdocs build system. This will help prevent that in the future.

PR #609, opened 2020-02-28
MQE-1057: Update metadata filenames convention

tomreece:
Need to update technical guidelines.

Related PRs to MFTF-3.0.0 develop branches:
* https://github.com/magento-pangolin/magento2ce/pull/234
* https://github.com/magento-pangolin/magento2ee/pull/109
* https://github.com/magento-pangolin/magento2b2b/pull/38
* https://github.com/magento-pangolin/inventory/pull/1
* <note: no pagebuilder changes required>
* Technical guidelines: https://github.com/magento/devdocs/pull/6764

Build
https://m2build-ur.devops.magento.com/job/All-User-Requested-Tests/30315/

The above build has failures, but they are equivalent to the MFTF-3.0.0 branch failures.


tomreece:
> Could see a couple more usages of -meta.xml in mftf. Can you please check if they need to be renamed as well -

Those files are outdated and unused. I did not touch these files because they should be removed from our framework. If I fix the comment, I have to fix and test the implementation of those which generate `.xml` files and we never test that code.

Is it ok if I leave it as-is?

PR #668, opened 2020-03-31
MQE-2042: Port MFTF Coverage to Jenkins

tomreece:
Modify TestContextExtension to call test.php file to keep track of test name during code coverage gathering.
Controlled by a new environment variable called `ENABLE_CODE_COVERAGE`.

### DIFF
--- a/etc/codecoverage/index.php
+++ b/etc/codecoverage/index.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * Application entry point
+ *
+ * Example - run a particular store or website:
+ * --------------------------------------------
+ * require __DIR__ . '/app/bootstrap.php';
+ * $params = $_SERVER;
+ * $params[\Magento\Store\Model\StoreManager::PARAM_RUN_CODE] = 'website2';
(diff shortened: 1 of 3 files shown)


etc/codecoverage/index.php
```diff
@@ -0,0 +1,58 @@
...
+ * $params[\Magento\Store\Model\StoreManager::PARAM_RUN_TYPE] = 'website';
+ * $bootstrap = \Magento\Framework\App\Bootstrap::create(BP, $params);
+ * \/** @var \Magento\Framework\App\Http $app *\/
+ * $app = $bootstrap->createApplication(\Magento\Framework\App\Http::class);
+ * $bootstrap->run($app);
+ * --------------------------------------------
+ *
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
```
jilu1, line 19: Can we move copyright to top of file?

tomreece: This is where it's at in the original index.php from magento2ce.
I don't want to change it and introduce an unnecessary diff.
It matches CE.

jilu1: Ok. I did not know that

PR #676, opened 2020-04-17
MQE-2086: Reduce amount of attachments in Allure reports

KevinBKozan:
### Description
* New `.env` flag for `VERBOSE_ARTIFACTS
* Docs updates

### Contribution checklist


dev/tests/functional/standalone_bootstrap.php
```diff
@@ -53,6 +53,9 @@
     defined('WAIT_TIMEOUT') || define('WAIT_TIMEOUT', 30);
     $env->setEnvironmentVariable('WAIT_TIMEOUT', WAIT_TIMEOUT);
 
+    defined('VERBOSE_ARTIFACTS') || define('VERBOSE_ARTIFACTS', false);
```
jilu1, line 56: The logic of VERBOSE_ARTIFACTS flag is not quite right. Maybe you should use `string` instead of `boolean`.

src/Magento/FunctionalTestingFramework/Allure/Adapter/MagentoAllureAdapter.php
```diff
@@ -354,4 +359,21 @@ private function retrieveStepKey($stepLine)
...
+    /**
+     * Removes attachments from step depending on MFTF configuration
+     * @param Step    $step
+     * @param Failure $testFailed
+     * @return void
+     */
+    private function removeAttachments($step, $testFailed)
+    {
+        //Remove Attachments if verbose flag is not true AND test did not fail
+        if (getenv('VERBOSE_ARTIFACTS') !== true && $testFailed === null) {
```
jilu1, line 372: Please check this condition again after address earlier comment.

dev/tests/functional/standalone_bootstrap.php
KevinBKozan: Hm, where does this get weird, in the comparisons in the file below? What behavior are you seeing?

jilu1: I found the flag is not working as expected and it turns out VERBOSE_ARTIFACTS is empty string.

KevinBKozan: I see that, only happens if the flag is not present at all. I fixed this by nakign it set `false` instead, and changed the string comparison in the below file.

src/Magento/FunctionalTestingFramework/Allure/Adapter/MagentoAllureAdapter.php
KevinBKozan: Fixed

PR #697, opened 2020-05-07
MQE-2047: Jenkins Pipeline - Static Check Options

KevinBKozan:
### Description
* static checks now check for existance of a `staticRuleset.json` under wither `dev/tests/acceptance` or `dev` in case of embedded/standalone respectively.

### Contribution checklist


src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
```diff
@@ -81,6 +88,7 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->ioStyle = new SymfonyStyle($input, $output);
+        $this->parseRulesetJson();
```
soumyau, line 91: ruleset should not be read if an argument is passed to the static-check command, correct?

Currently I see this:
```
Soumyas-MBP:magento2-functional-testing-framework sunnikri$ bin/mftf static-checks deprecatedEntityUsage
Using ruleset under /Users/sunnikri/workspace/magento2-functional-testing-framework/dev/staticRuleset.json

Running static check script for: deprecatedEntityUsage
```

dev/staticRuleset.json
```diff
@@ -0,0 +1,6 @@
+{
```
soumyau, line 1: should this file even be present in MFTF? Even for standalone version, shouldn't MFTF read the file from magento dev/tests/acceptance path?

src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
```diff
@@ -164,4 +160,43 @@ private function validateInput(InputInterface $input)
...
+        }
+    }
+
+    /**
+     * Parses and sets local ruleSet. If not found, simply returns and lets script continue.
+     * @return void;
+     */
+    private function parseRulesetJson()
+    {
+        $pathToRuleset = TESTS_BP . DIRECTORY_SEPARATOR . "staticRuleset.json";
```
soumyau, line 194: should the file `staticRuleset.json` even be added to MFTF? Because, again updates to this file will need a new MFTF version. Reading from this path works for both standalone and embedded -
```
const DEV_TESTS_DIR = 'dev/tests/acceptance/';        
$devTestsPath = FilePathFormatter::format(MAGENTO_BP) . self::DEV_TESTS_DIR;
$pathToRuleset = $devTestsPath . DIRECTORY_SEPARATOR . "staticRuleset.json";
```

src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
```diff
@@ -164,4 +160,43 @@ private function validateInput(InputInterface $input)
...
+        if ($pathToRuleset === null) {
+            $this->ioStyle->text("No ruleset under $pathToRuleset" . PHP_EOL);
+            return;
+        }
+        $this->ioStyle->text("Using ruleset under $pathToRuleset" . PHP_EOL);
+        $this->ruleSet = json_decode(file_get_contents($pathToRuleset), true);
```
soumyau, line 200: the step throws a fatal error if file is not found. Good to have exception handled.

src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
KevinBKozan: I see what you mean. In the code it will use arguments over the ruleset, but it's misleading messaging. I'll fix it.

dev/staticRuleset.json
KevinBKozan: So we have a global `TESTS_BP` that changes depending on standalone vs embedded, and in embedded it's `dev/tests/acceptance`, that's why I opted to use it. Standalone will have TESTS_BP equal to `dev/tests/acceptance` if there's a defined `MAGENTO_BP`, so we shouldn't have an issue there.

As for if this file should be present in MFTF, I wanted to include it as an example. If you don't think we need one, then I'll happily get rid of it

src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
KevinBKozan: See other comment, TESTS_BP is set as following:
* If we detect we are attached to Magento (either by composer or MAGENTO_BP), it's `dev/tests/acceptance`
* If we are not attached, it's just `dev/` to point towards MFTF standalone `dev directory`

src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
KevinBKozan: Fair point, will fix.

dev/staticRuleset.json
KevinBKozan: Going to delete this file.

src/Magento/FunctionalTestingFramework/Console/StaticChecksCommand.php
soumyau: in standalone mode, with latest code, it tries to read `staticRuleSet.json` from mftf's `\dev`, doesn't find it, and executes all static checks instead. Shouldn't standalone also read from attached magento's `dev/tests/acceptance`?

However in case mftf is not attached at all i.e MAGENTO_BP is not set in `.env`, it again tries to read file from mftf's `\dev`, doesn't find it and executes all static checks, which seems correct.

soumyau: Also found another bug, though unrelated to this PR -
if MAGENTO_BP is not set in `.env` (detached mode) , deprecatedEntityUsage throws php errors. 
Simple fix would be to add an error message like `[ERROR] DEPRECATED ENTITY USAGE CHECK ABORTED: MFTF must be attached or pointing to Magento codebase.`

KevinBKozan: I think I understand better now, let me try and reproduce scenario 1.

PR #725, opened 2020-06-02
MQE-2140: [chrome81] dragAndDrop with x and y offsets no longer works

soumyau:

2. ...

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Module/MagentoWebDriver.php
+++ b/src/Magento/FunctionalTestingFramework/Module/MagentoWebDriver.php
@@ -745,20 +745,22 @@ class MagentoWebDriver extends WebDriver
      */
     public function dragAndDrop($source, $target, $xOffset = null, $yOffset = null)
     {
-        if ($xOffset !== null || $yOffset !== null) {
-            $snodes = $this->matchFirstOrFail($this->baseElement, $source);
-            $tnodes = $this->matchFirstOrFail($this->baseElement, $target);
+        $snodes = $this->matchFirstOrFail($this->baseElement, $source);
+        $tnodes = $this->matchFirstOrFail($this->baseElement, $target);
+        $action = new WebDriverActions($this->webDriver);
 
+        if ($xOffset !== null || $yOffset !== null) {
             $targetX = intval($tnodes->getLocation()->getX() + $xOffset);
             $targetY = intval($tnodes->getLocation()->getY() + $yOffset);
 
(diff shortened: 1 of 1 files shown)


KevinBKozan:
> 2. Please check with @KevinBKozan on why we need to use custom steps instead of the web driver drag and drop function in the first place. I remember it's some page builder requirements.
@jilu1 it was a pagebuilder requirement, yes. If you remove them and pagebuilder tests still pass then it should be fine.

src/Magento/FunctionalTestingFramework/Module/MagentoWebDriver.php
```diff
@@ -745,23 +745,24 @@ public function _before(TestInterface $test)
...
-            $action->moveByOffset($travelX, $travelY)->perform();
+            $action->moveToElement($snodes);
+            $action->clickAndHold($snodes);
+            $action->moveByOffset($travelX, $travelY);
             $action->release()->perform();
         } else {
-            parent::dragAndDrop($source, $target);
+            $action->clickAndHold($snodes);
+            $action->moveToElement($tnodes);
+            $action->release($tnodes)->perform();
```
jilu1, line 765: Any reason for this change? I remember this issue only applies for dragAndDrop with offset.

soumyau: for chrome 81, separating out dragAndDrop into it's native functions helped resolve couple of issues with drag and drop without offsets too, hence the fix.

soumyau:
> 1. Changes look good. I only have one question.
> 2. Please rebase your branch. There are too many revision of the MagentoWebDriver.php in your branch.
Will do 👍 
> 3. From the test fix, it looks like instead of calculating x and y offset, we now use -1. Can we document how user can calculate this value? Will it always be -1?
It's won't be always one, there are tests which use calculated offsets.
> 4. Pull latest mainline and run builds with all required extensions to make sure that fix works in all dragAndDrop tests for Chrome65 and Chrome81 if you have not done so.
I'll trigger another build and add link here 👍

src/Magento/FunctionalTestingFramework/Module/MagentoWebDriver.php
jilu1: Interesting to know. Maybe php webdriver need to make the some change?

soumyau: yes, we are on latest version of php webdriver. It's probably page builder specific.

PR #772, opened 2020-07-28
MQE-1974: Report used deprecated metadata in Test

soumyau:
Changes -
1. Added logging capability in OperationDefinitionObject for used deprecated operation in a test. ( logged at runtime)
2. Updated MftfLogger to log all warnings and deprecations during test generation and runtime. This behavior was previously changed to log only during test generation to make mftf.log smaller. All Operation classes are initialized only during runtime; so these validations happen at runtime.
3. Name validations for DataObjectHandler occur twice during test generation and test runs - changed it to once. This ensures all logs in mftf.log are unique.
4. Modified deprecation warning messages across all entities

2. ...

### Contribution checklist

### DIFF
--- a/src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/DataObjectHandler.php
+++ b/src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/DataObjectHandler.php
@@ -75,7 +75,7 @@ class DataObjectHandler implements ObjectHandlerInterface
 
     /**
      * Constructor
-     * @param bool $validateName Set to false to disable name validations
+     * @param boolean $validateName Set to false to disable name validations.
      * @throws XmlException
      */
     private function __construct($validateName = true)
@@ -94,7 +94,7 @@ class DataObjectHandler implements ObjectHandlerInterface
     /**
      * Return the singleton instance of this class. Initialize it if needed.
      *
-     * @param bool $validateName
+     * @param boolean $validateName
      * @return DataObjectHandler
      * @throws \Exception
      */
@@ -138,10 +138,11 @@ class DataObjectHandler implements ObjectHandlerInterface
      * Convert the parser output into a collection of EntityDataObjects
      *
      * @param string[] $parserOutput Primitive array output from the Magento parser.
-     * @param bool     $validateName
+     * @param boolean  $validateName
      * @return EntityDataObject[]
      * @throws XmlException
      * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD.NPathComplexity)
      */
     private function processParserOutput($parserOutput, $validateName = true)
     {
@@ -245,7 +246,7 @@ class DataObjectHandler implements ObjectHandlerInterface
      * Parses <data> elements in an entity, and returns them as an array of "lowerKey"=>value.
      *
      * @param string[] $entityData
-     * @param bool     $validateName
+     * @param boolean  $validateName
      * @return string[]
      */
     private function processDataElements($entityData, $validateName = true)
--- a/src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/DataObjectHandler.php
+++ b/src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/DataObjectHandler.php
@@ -75,10 +75,8 @@ class DataObjectHandler implements ObjectHandlerInterface
 
     /**
      * Constructor
-     * @param boolean $validateName Set to false to disable name validations.
-     * @throws XmlException
      */
-    private function __construct($validateName = true)
+    private function __construct()
     {
         $parser = ObjectManagerFactory::getObjectManager()->create(DataProfileSchemaParser::class);
         $parserOutput = $parser->readDataProfiles();
@@ -87,21 +85,20 @@ class DataObjectHandler implements ObjectHandlerInterface
         }
         $this->entityNameValidator = new NameValidationUtil();
         $this->entityKeyValidator = new NameValidationUtil();
-        $this->entityDataObjects = $this->processParserOutput($parserOutput, $validateName);
+        $this->entityDataObjects = $this->processParserOutput($parserOutput);
         $this->extendUtil = new DataExtensionUtil();
     }
 
     /**
      * Return the singleton instance of this class. Initialize it if needed.
      *
-     * @param boolean $validateName
      * @return DataObjectHandler
      * @throws \Exception
      */
-    public static function getInstance($validateName = true)
+    public static function getInstance()
     {
         if (!self::$INSTANCE) {
-            self::$INSTANCE = new DataObjectHandler($validateName);
+            self::$INSTANCE = new DataObjectHandler();
         }
         return self::$INSTANCE;
     }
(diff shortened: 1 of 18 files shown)


src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/DataObjectHandler.php
```diff
@@ -135,11 +138,12 @@ public function getAllObjects()
      * Convert the parser output into a collection of EntityDataObjects
      *
      * @param string[] $parserOutput Primitive array output from the Magento parser.
+     * @param boolean  $validateName
      * @return EntityDataObject[]
      * @throws XmlException
-     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD)
```
jilu1, line 144: Can we suppress individual PHPMD warnings? It helps us know the potential problem and refactor in the future

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/PersistedObjectHandler.php
```diff
@@ -87,7 +87,7 @@ public function createEntity(
             $retrievedDependentObjects[] = $this->retrieveEntity($objectKey, $scope);
         }
         
-        $retrievedEntity = DataObjectHandler::getInstance()->getObject($entity);
+        $retrievedEntity = DataObjectHandler::getInstance(false)->getObject($entity);
```
jilu1, line 90: DataObjectHandler::getInstance() is used everywhere in our code. What's the rule to set the flag to true or false? Why set it to false here?

src/Magento/FunctionalTestingFramework/DataGenerator/Persist/DataPersistenceHandler.php
```diff
@@ -110,7 +110,7 @@ public function updateEntity($updateDataName, $updateDependentObjects = [])
         foreach ($updateDependentObjects as $dependentObject) {
             $this->dependentObjects[] = $dependentObject->getCreatedObject();
         }
-        $updateEntityObject = DataObjectHandler::getInstance()->getObject($updateDataName);
+        $updateEntityObject = DataObjectHandler::getInstance(false)->getObject($updateDataName);
```
jilu1, line 113: Why here the flag should be `false`?

src/Magento/FunctionalTestingFramework/Util/Logger/MftfLogger.php
```diff
@@ -86,13 +81,10 @@ public function criticalFailure($message, array $context = [], $verbose = false)
     public function notification($message, array $context = [], $verbose = false)
     {
         $message = "NOTICE: " . $message;
-        // print during test generation
-        if ($this->phase === MftfApplicationConfig::GENERATION_PHASE && $verbose) {
+        // Suppress print during unit testing
+        if ($this->phase !== MftfApplicationConfig::UNIT_TEST_PHASE && $verbose) {
```
jilu1, line 85: Can we safely revert the code recently done for trimming the log file size?

src/Magento/FunctionalTestingFramework/Util/Logger/MftfLogger.php
```diff
@@ -86,13 +81,10 @@ public function criticalFailure($message, array $context = [], $verbose = false)
...
             print ($message . json_encode($context) . "\n");
         }
-        // suppress logging during test execution
-        if ($this->phase !== MftfApplicationConfig::EXECUTION_PHASE) {
-            parent::notice($message, $context);
-        }
+        parent::notice($message, $context);
```
jilu1, line 88: Can we safely revert the code recently done for trimming the log file size?

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/PersistedObjectHandler.php
soumyau: suppressed name validation while calling `createEntity`, `getEntity`, `updateEntity`. These functions are only invoked during runtime. Looks DataObjectHandler, creates a new instance of the class during runtime which is why these validations happen again (not 100% sure, that's what it looks like). Rule probably is - suppress during runtime instantiation.

src/Magento/FunctionalTestingFramework/Util/Logger/MftfLogger.php
soumyau: Um, we could. There is some code left behind from that PR like use of `$this->phase`. I figured, updating code would be easier, given that we'll need to delete history etc. Let me know otherwise.

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/DataObjectHandler.php
soumyau: fixed

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/PersistedObjectHandler.php
jilu1: It's not necessary to use a new flag. Let's just use `MftfApplicationConfig::EXECUTION_PHASE`

src/Magento/FunctionalTestingFramework/Util/Logger/MftfLogger.php
jilu1: I am not clear what you mean. This change results in printing/logging notices, doesn't it?

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/PersistedObjectHandler.php
jilu1: Adding this flag in widely used framework code to avoid displaying deprecation message more than once seems not unnecessary. I agree with @okolesnyk, if we cannot avoid displaying on generation time and runtime for deprecated metadata, it's fine to display twice.

src/Magento/FunctionalTestingFramework/Util/Logger/MftfLogger.php
jilu1: I would prefer keep the original behavior: do not print/log notices during runtime. One possible solution might be to use $context. If it's metadata deprecation notice, print/log it, else skip the notices

soumyau: fixed. Now using $context to print/log deprecation notices during runtime.

soumyau:
> Please add verification test according to AC

done

src/Magento/FunctionalTestingFramework/DataGenerator/Handlers/PersistedObjectHandler.php
soumyau: Removed flag as per comments.

src/Magento/FunctionalTestingFramework/Util/Logger/MftfLogger.php
soumyau: reverted changes

soumyau:
> It looks good! Have you rerun the UR builds after the change?

rerun build - looks good.

PR #775, opened 2020-07-29
MQE-2229: Deprecation Error When Deprecated ActionGroup References De…

soumyau:
…precated Element

2. ...

### Contribution checklist

### DIFF
--- a/dev/tests/unit/Util/ObjectHandlerUtil.php
+++ b/dev/tests/unit/Util/ObjectHandlerUtil.php
@@ -6,45 +6,31 @@
 namespace tests\unit\Util;
 
 use AspectMock\Test as AspectMock;
+use Magento\FunctionalTestingFramework\DataGenerator\Handlers\DataObjectHandler;
 use Magento\FunctionalTestingFramework\DataGenerator\Handlers\OperationDefinitionObjectHandler;
+use Magento\FunctionalTestingFramework\DataGenerator\Parsers\DataProfileSchemaParser;
 use Magento\FunctionalTestingFramework\DataGenerator\Parsers\OperationDefinitionParser;
 use Magento\FunctionalTestingFramework\ObjectManager;
 use Magento\FunctionalTestingFramework\ObjectManagerFactory;
+use Magento\FunctionalTestingFramework\Page\Handlers\PageObjectHandler;
+use Magento\FunctionalTestingFramework\Page\Handlers\SectionObjectHandler;
+use Magento\FunctionalTestingFramework\Test\Handlers\ActionGroupObjectHandler;
+use Magento\FunctionalTestingFramework\Test\Handlers\TestObjectHandler;
+use Magento\FunctionalTestingFramework\Test\Parsers\ActionGroupDataParser;
+use Magento\FunctionalTestingFramework\Test\Parsers\TestDataParser;
+use Magento\FunctionalTestingFramework\XmlParser\PageParser;
+use Magento\FunctionalTestingFramework\XmlParser\SectionParser;
 
 class ObjectHandlerUtil
 {
     /**
-     * @var ObjectHandlerUtil
-     */
-    private static $instance;
-
-    /**
-     * ObjectHandlerUtil constructor.
-     */
-    private function __construct()
-    {
-        // private constructor
-    }
-    /**
-     * Static singleton get function
-     *
-     * @return ObjectHandlerUtil
-     */
-    public static function getInstance()
-    {
-        if (self::$instance == null) {
-            self::$instance = new ObjectHandlerUtil();
-        }
-
-        return self::$instance;
-    }
-    /**
-     * Function used to set mock for Operation parser.
+     * Set up everything required to mock OperationDefinitionObjectHandler::getInstance() with $data value
      * @param array $data
      * @throws \Exception
      */
-    public function setMockOperationParserOutput($data)
+    public static function mockOperationHandlerWithData($data)
     {
+        // Clear OperationDefinitionObjectHandler singleton if already set
         $property = new \ReflectionProperty(
             OperationDefinitionObjectHandler::class,
             'INSTANCE'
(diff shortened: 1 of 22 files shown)


soumyau:
> * We should add this fix to data and metadata.
>   Good catch, fixed.
> * There is no unit test before and I remember i ask if unit test is required, i think we decided to add it. Let us know if you have concern.
>   Not sure how to write a unit test for a static-check by mocking file system, InputInterface etc. It may not even be possible to write granular unit tests for this command.

soumyau:
@KevinBKozan re: https://jira.corp.magento.com/browse/MQE-2229. AC states to add unit tests. 

> > * We should add this fix to data and metadata.
> >   Good catch, fixed.
> > * There is no unit test before and I remember i ask if unit test is required, i think we decided to add it. Let us know if you have concern.
> >   Not sure how to write a unit test for a static-check by mocking file system, InputInterface etc. It may not even be possible to write granular unit tests for this command.
@KevinBKozan : https://jira.corp.magento.com/browse/MQE-2229 states to add unit tests. Let us know if if you agree with the above.

KevinBKozan:
@soumyau yeah unit testing this is a little involved. You'd need to:
* Mock call to `$this->scriptUtil->printErrorsToFile()` (easy)
* Mock `loadAllXmlFiles()` to populate all of the `$this->xml` variables (easy)
Then in the actual function we're testing we have to:
* Abstract `file_get_contents() => $this->loadFile()` so we can mock it to return xml string + change call in line to `$domDocument->loadXML($contents);` (easy)
* Mock all required `*Handler` classes like https://github.com/magento/magento2-functional-testing-framework/blob/MQE-2229/dev/tests/unit/Magento/FunctionalTestFramework/Test/Objects/ActionObjectTest.php#L377 (tricky but simple, need to be well organized to stay sane)
* Doing above then allows scriptUtil to resolve them easily with no mocking required in scriptUtil due to the logic in https://github.com/magento/magento2-functional-testing-framework/blob/MQE-2229/src/Magento/FunctionalTestingFramework/Util/Script/ScriptUtil.php#L280

Personally I think we *should* add Unit tests, if only because we're not in a time crunch. Also the first class to get unit tests in an area of the framework is always the most arduous, I think it's a solid opportunity to add a lot of coverage to an area of the framework that is pretty barren right now.

If you want to meet in the middle: timebox writing unit tests to 1 day. If it ends up being much trickier than the above we can make stories to add coverage to this part of the FW. Does that sound fair?

soumyau:
> @soumyau yeah unit testing this is a little involved. You'd need to:
> 
> * Mock call to `$this->scriptUtil->printErrorsToFile()` (easy)
> * Mock `loadAllXmlFiles()` to populate all of the `$this->xml` variables (easy)
> 
> Then in the actual function we're testing we have to:
> 
> * Abstract `file_get_contents() => $this->loadFile()` so we can mock it to return xml string + change call in line to `$domDocument->loadXML($contents);` (easy)
> * Mock all required `*Handler` classes like https://github.com/magento/magento2-functional-testing-framework/blob/MQE-2229/dev/tests/unit/Magento/FunctionalTestFramework/Test/Objects/ActionObjectTest.php#L377 (tricky but simple, need to be well organized to stay sane)
> * Doing above then allows scriptUtil to resolve them easily with no mocking required in scriptUtil due to the logic in https://github.com/magento/magento2-functional-testing-framework/blob/MQE-2229/src/Magento/FunctionalTestingFramework/Util/Script/ScriptUtil.php#L280
> 
> Personally I think we _should_ add Unit tests, if only because we're not in a time crunch. Also the first class to get unit tests in an area of the framework is always the most arduous, I think it's a solid opportunity to add a lot of coverage to an area of the framework that is pretty barren right now.
> 
> If you want to meet in the middle: timebox writing unit tests to 1 day. If it ends up being much trickier than the above we can make stories to add coverage to this part of the FW. Does that sound fair?

@KevinBKozan sure thing. Thanks for the direction. I'll give this a (timeboxed) shot, if the effort is larger, I'll create follow up tickets.

dev/tests/unit/Util/ObjectHandlerUtil.php
```diff
@@ -0,0 +1,73 @@
...
+    private function __construct()
+    {
+        // private constructor
+    }
+    /**
+     * Static singleton get function
+     *
+     * @return ObjectHandlerUtil
+     */
+    public static function getInstance()
```
KevinBKozan, line 33: See what you're trying to accomplish with this class, and I think it's a good idea, but I don't think we should introduce a new singleton. Since this class doesn't keep track of any data, we should refrain from introducing a singleton.

This class should instead be a static class instead with static functions `setMock<type>HandlerWithData($data)`.just like you have it, with a cleanup helper function as well. I confirmed that it works locally in your new testcases.

dev/tests/unit/Util/ObjectHandlerUtil.php
```diff
@@ -0,0 +1,73 @@
...
+        AspectMock::double(ObjectManagerFactory::class, ['getObjectManager' => $instance]);
+    }
+
+    /**
+     * Function which clears the mock handler context from the ObjectHandlerUtil class.
+     * Should be run after a test class has executed.
+     *
+     * @return void
+     */
+    public function clearMockObjectHandlerUtil()
```
KevinBKozan, line 69: This is never called; also, iI believe t should clean the `Handler` classes, not this util class. The handlers should be the ones that contain the mocked `INSTANCE`.

KevinBKozan:
To test `loadAllXmlFiles` we have two approaches we can take:
* Mock scriptUtil to death
* Introduce a fake module under `unit/Resources/DeprecatedUsageModules/` with folder `unit/Resources/DeprecatedUsageModules/TestLoadXML/Test/` `unit/Resources/DeprecatedUsageModules/TestLoadXML/Suite` `unit/Resources/DeprecatedUsageModules/TestLoadXML/ActionGroup`, etc for each function.

Option two is honestly more of an integration test, but it makes far more sense from a value standpoint than endlessly mocking scriptUtil. A huge part of the value of writing the test is knowing that xml files are actually loaded for use in other functions. So what I think:
* under `verification/Tests`, introduce `DeprecationCheckTest` to cover this specific usecase
* create a new `TestModule` copy like `DeprecationCheckModule` or `DeprecationCheckModules/<specificUseCaseModule>/` with all the correct subfolders to support the testcase so that the test is *actually* reading real files and testing that functionality.

dev/tests/unit/Util/ObjectHandlerUtil.php
soumyau: fixed 👍

dev/tests/unit/Util/ObjectHandlerUtil.php
soumyau: fixed 👍

soumyau:
> Small comment on singleton usage; overall works looks solid, just need minor touchup.

Fixed 👍

soumyau:
> Small comment on singleton usage; overall works looks solid, just need minor touchup.

Done.

PR #779, opened 2020-07-31
MQE-2158: support using actions from multiple modules in Suites

jilu1:

2. ...

### Contribution checklist

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -24,8 +24,16 @@ If there is no response from the contributor for two weeks, the issue is closed.
 Often when the MFTF team works on reviewing the suggested changes, we will add a label to the issue to indicate to our internal team certain information, like status or who is working the issue.
 If you’re ever curious what the different labels mean, see the [table][labels] below for an explanation of each one.
 
-Refer to [Magento Contributor Agreement] for detailed information about the License Agreement.
-All contributors are required to submit a click-through form to agree to the terms.
+## Code Of Conduct
+
+This project adheres to the Adobe [Code Of Conduct](../CODE_OF_CONDUCT.md). By participating, you are expected to uphold this code. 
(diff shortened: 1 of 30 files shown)


soumyau:
changes look good. Had a couple of questions -
1.  What is the reasoning behind accessing module container from MagentoWebDriver? Can it live in it's own class?
2. How do I use actions outside of MagentoWebDriver in a Helper?

jilu1:
> changes look good. Had a couple of questions -
> 
> 1. What is the reasoning behind accessing actions in other custom modules from MagentoWebDriver? Can it live in a parent class that all other custom modules extend?
An alternative solution is use a base module that all our modules extend but I think that's too overkill. MagentoWebDriver is for sure used in Suite to handling sessions.
> 2. How do I use actions outside of MagentoWebDriver in a Helper? Can that use case be included in a verification test?
I think helpers can access anything if they want to. Tests to show how helps use actions from other modules if needed is not in scope of this story.

soumyau:
> > changes look good. Had a couple of questions -
> > 
> > 1. What is the reasoning behind accessing actions in other custom modules from MagentoWebDriver? Can it live in a parent class that all other custom modules extend?
> 
> An alternative solution is use a base module that all our modules extend but I think that's too overkill. 
MagentoWebDriver is for sure used in Suite to handling sessions.
> Sounds good.
> > 1. How do I use actions outside of MagentoWebDriver in a Helper? Can that use case be included in a verification test?
> 
> I think helpers can access anything if they want to. Tests to show how helps use actions from other modules if needed is not in scope of this story.
looks like `$this->getModule('\Magento\FunctionalTestingFramework\Module\MagentoWebDriver')` is used in Helper classes and suites. Since actions are moved out of this class, maybe a similar mechanism of accessing container will be useful in case of Helpers too.

PR #799, opened 2020-09-17
MQE-2289: Cant generate urn catalog

tomreece:
Fixes documentation to have correct `generate:urn-catalog` examples.


KevinBKozan:
@tomreece can confirm this is required for the command to work.

KevinBKozan (approved):

PR #904, opened 2023-05-30
Wait for cron process to finish properly

fredden:
### Description
When running `bin/magento cron:run`, several background processes are spawned. The intent of the `<magentoCron />` XML node is to run certain cron groups to completion. However, as the actual processing may happen in a background process, the test-suite typically does not behave as expected. (This is a race condition; see https://github.com/magento/magento2/pull/37511 for further details.)

This pull request solves this race condition by using the internal 'bootstrap' argument to indicate that the cron process is already standalone. This means that no additional background processes will be run and the jobs complete in the foreground process. The test-suite will then wait for this process to terminate before proceeding. This is the desired behaviour.

### Contribution checklist
 - [x] All automated tests passed successfully (all builds are green)

### DIFF
--- a/src/Magento/FunctionalTestingFramework/Module/MagentoWebDriver.php
+++ b/src/Magento/FunctionalTestingFramework/Module/MagentoWebDriver.php
@@ -1033,6 +1033,10 @@ class MagentoWebDriver extends WebDriver
     {
         $cronGroups = array_filter($cronGroups);
 
+        if (isset($cronGroups[0]) && !isset($cronGroups[1])) {
+            $arguments .= ' --bootstrap=standaloneProcessStarted=1';
+        }
+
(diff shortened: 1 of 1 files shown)


fredden:
Please don't let any test failures in https://github.com/magento/magento2/pull/37511 block this pull request. It's stand-alone on purpose.

KevinBKozan:
@fredden Thank you for your contribution! We merged your commits in an internal PR after testing the functionality changes. This PR should automatically close once they are synced 👍

fredden:
@KevinBKozan thanks for the update. Do you know when a new release will be tagged which includes this change?

KevinBKozan:
@fredden I believe it should be *very* soon, a day or two at most. Internal PR to finalize the version was already merged so the new tag should be out soon!

Issue #915, opened 2024-01-12
Console command build:project starts subprocess with system-default PHP instead of the runtime PHP


kanevbg:
Running the console command `build:project` with non-default PHP is not working as expected because the command runs sub-process `vendor/bin/codecept build` without specifying the PHP binary to run with, that in results makes the composer auto-loader utilize the default PHP binary from the system, which is not expected as we may have executed the `build:project` command with non system default PHP, eg:
```
php81 vendor/bin/mftf build:project
```

### Preconditions
No preconditions really, it's logical issues, but to see a practical problem you may use PHP 7.4 (or older) as default  system PHP and install the MFTF with PHP 8.1 binary not aliased as `php` in the bash environment.

### Expected result
Successful build.

### Actual result
```
Fatal error: Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.1.0". You are running 7.4.29. in vendor/composer/platform_check.php on line 24

In BuildProjectCommand.php line 106:
                                                                                                 
  The codecept build command failed unexpectedly. Please see the above output for more details.                                                                                                
```

PR #916, opened 2024-01-12
Fix for issue 915

kanevbg:
Fixes issue #915


kanevbg:
Update: that may be not proper solution. I have found that something else is wrong on 3 more places a couple of days after making that PR and applied some other change, but haven't had time to update the PR and information. Just commenting to let you know. The actual fix may be other.

